PR #36, opened 2021-04-08
Feature/build integration

n0mn0m:
Update API calls and route integration.

This is working with Django locally. Get some conflicts with the APIs in blackbalsam deployment that I'm going to troubleshoot seperately.


n0mn0m:
@connectbo one note `DELETE` doesn't work. `app_id` is undefined, but I haven't had time to dig down into that yet.

connectbo:
@n0mn0m Thanks for those! I will put that `DELETE` function on my list, and then pr.

PR #41, opened 2021-04-09
LTS update

n0mn0m:
Update to node 14 for LTS support.

Tested with compose, and pushed an image then tested with appstore. Artifacts from react built with node 14 work as expected.

https://nodejs.org/en/about/releases/

ref: helxplatform/development#700


connectbo (approved):

n0mn0m:
@connectbo going to hold this till the CI PR is worked early next week so CI runs with node 14 too.

PR #43, opened 2021-04-13
Feature/search study

connectbo:
This pr features:
- display study variables between search results and knowledge graph (same as dug search)
- display no results when receiving empty response
- clean up log info

I tried to dig into the last week's rendering issue on search page where query is updated but results aren't. I haven't encountered that issue so far, but will look into that once occur.

### DIFF
--- a/src/components/search/search-context.js
+++ b/src/components/search/search-context.js
@@ -115,6 +115,21 @@ export const HelxSearch = ({ children }) => {
     return knowledgeGraphs.map(graph => graph._source.knowledge_graph.knowledge_graph)
   }
 
+  const fetchStudyVariable = async (_id, _query) => {
+    const studyVariables = await axios.post(`${helxSearchUrl}/search_var`, {
+      concept: _id,
+      index: 'variables_index',
+      query: _query,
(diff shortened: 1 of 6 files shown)


src/components/search/search-context.js
```diff
@@ -115,6 +115,21 @@ export const HelxSearch = ({ children }) => {
...
+  const fetchStudyVariable = async (_id, _query) => {
+    const studyVariables = await axios.post(`${helxSearchUrl}/search_var`, {
+      concept: _id,
+      index: 'variables_index',
+      query: _query,
+      size: 1000
+    }).then(response => {
+      return response.data.result.hits.hits
+    }).catch(error => {
+      console.error(error)
```
n0mn0m, line 127: Would like to have a place to route these one day.

n0mn0m (approved):
Looks good!

src/components/search/search-context.js
connectbo: Yeah, I think my next update will add more notification UI when user interaction happens. That could go to the error notification :)

PR #46, opened 2021-04-15
Feature/app notifications

connectbo:
This pr features
- Axios global error interceptor to handle 403 issue and redirect to /frontend
- Change search page to be homepage
- Page will reload data when an instance is stopped
- Add notifications when launching and stopping instances.

Let me know if these works, can hop on a call if necessary. @n0mn0m

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -23,9 +23,10 @@ const App = () => {
             <LocationProvider>
               <Layout>
                 <Router>
-                  <Home path="/frontend/react" />
+                  <Home path="/frontend/react/home" />
                   <Apps path="/frontend/react/apps" />
                   <Account path="/frontend/react/account" />
+                  <Search path="/frontend/react" />
                   <Search path="/frontend/react/search/*" />
                   <Branding path="/frontend/react/branding" />
                   <NotFound default />
--- a/src/components/search/search-form.js
+++ b/src/components/search/search-form.js
@@ -45,7 +45,7 @@ export const SearchForm = () => {
   return (
     <SearchBarContainer>
       <SearchBarInputGroup>
-        <Input ref={inputRef} value={searchTerm} onChange={handleChangeQuery} onKeyDown={handleKeyDown} style={{ flex: 1 }} />
+        <Input placeholder="Search HeLx" ref={inputRef} value={searchTerm} onChange={handleChangeQuery} onKeyDown={handleKeyDown} style={{ flex: 1 }} />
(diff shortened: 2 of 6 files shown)


src/app.js
```diff
@@ -23,9 +23,10 @@ const App = () => {
             <LocationProvider>
               <Layout>
                 <Router>
-                  <Home path="/frontend/react" />
+                  <Home path="/frontend/react/home" />
```
n0mn0m, line 26: Nit, what are your thoughts on calling what we currently have as `home` about? We don't really have a "home" page/component since the default page is search and the information rendered here is more of an about style snippet?

n0mn0m (commented):
Overall looks good. Pulling to test locally and I'll let you know if it all works and looks good on my end.

src/app.js
connectbo: Good question. My initial thought is we can call the previous homepage 'About HeLx' page and change menu name as well. Unsure if that's appropriate, but can discuss that tomorrow

n0mn0m:
I get this when loading the frontend for the first time in appstore

image

That then seems to disrupt the apps loading

image

Also I think we should center the `no apps` and `no instances running` messages in the container.

One last thought and we can talk about this on the scrum tomorrow, but since search is the first page a user sees the `home` button navigating to a page they haven't seen feels strange. I would personally expect the `home` button to take me to the first page I see on login, but that may just be my personal expectation.

connectbo:
looking into those, will keep you updated.

n0mn0m:
> looking into those, will keep you updated.

Btw, was able to generate the same behavior with `make build compose`.

connectbo:
Give it a try. Axios calls should work now.

n0mn0m:
Updates look good. Would it be worth delaying the render of the `no apps` and `no active instances` by a few milliseconds or second? Currently when I toggle them I briefly see the message, then it flashes away as we render the app list or instance list.

connectbo:
Good point. I haven't tested this case on my machine, but I have some ideas. Fixing on the way :)

n0mn0m:
I noticed it with `make build compose` just toggling the two tabs if that helps.

connectbo:
Ok. The container below tabs on /apps page will now stay blank until the response is received, and then depending on the response, it would render either the empty app/instance message or entire listing.

n0mn0m (approved):
Works and looks great! Thanks for the updates.

PR #50, opened 2021-04-21
Slider memory units

n0mn0m:
@connectbo I wanted to push this up in a draft to get your ideas around the memory slider.

If you run this branch the changes you will see are:

- The conversion function handles the memory values
- Update displayed min max

I spent some time thinking about this, and because the compose spec supports multiple unit types (as does [appstore](https://github.com/helxplatform/appstore/blob/master/appstore/api/v1/validators.py) and [tycho](https://github.com/helxplatform/tycho/blob/master/tycho/client.py)) I think that one of the things we want to do is to convert to bytes, then operate to get a decent human readable interpretation. That interpretation can also be posted back to the appstore and tycho. Let me now your thoughts on that.

One place this causes trouble is the slider. We don't want to lose the metric type, but [range](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#step) wants to operate on a numeric. I think we could stuff the unit type into a prop, but figured I would see what you think too.

While working on this I also noticed that the way we set currentMemory, currentCpu and currentGpu in local storage. This means that if the app min values change between the last and current visit, but localStorage has not been cleared then we will see stale values here. I ran into this while working on the memory and wondering why I wasn't see the memory slider default value render with a metric till I manually changed localStorage between builds. Should this be handled via props or state and not in the browser global state? Since the names of the localStorage values are not app specific it seem there is room for name/value collisions and surprises too? 

Let me know what you think and the ideas you have around this. I'm going to switch over to the logout item on #707 as I know you are working on the layout right now too.

### DIFF
--- a/src/views/apps.js
+++ b/src/views/apps.js
@@ -136,6 +136,36 @@ const SliderMinMaxContainer = styled.span`
   width: 50vw;
 `
 
+function bytePower(unit_type) {
+  let ut = unit_type.toLowerCase()
+  const power = {
+    'k': 3,
+    'm': 6,
+    'g': 9,
+    't': 12,
+    'p': 15,
+    'e': 18,
+    'z': 21,
+    'y': 24,
+  }
+
+  return power[ut];
+}
+function formatBytes(mb, decimals = 2) {
+  let units = mb.substring(0, mb.length - 1)
+  let power_value = bytePower(mb[mb.length-1])
+  let bytes = units * Math.pow(10, power_value);
+
+  const k = 1000;
+  const dm = decimals < 0 ? 0 : decimals;
+  const sizes = ['B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'];
+
+  const i = Math.floor(Math.log(bytes) / Math.log(k));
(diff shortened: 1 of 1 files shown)


connectbo:
@n0mn0m Thanks for bringing this up! My thoughts would be
- Yes I think converting to bytes makes sense to me. In that way we can always have the precise value. I don't have any concern on stuffing the unit type into a prop, let's try that!
- For the localstorage part, I found out those as well when refactoring the apps page yesterday. There are some cases might cause error, one of them is when the min values changes and current value is not in its range. In the past, I think Steve mentioned we should have the ability to save the config specs when apps are launched, so that's why localstorage/sessionstorage is used here (we don't want to lose those when a page refresh, state variable is good but would be cleared on reload) But those haven't been tested on a live appstore side. I have some ideas to solve this, and can include it among the upcoming new apps page pr.

n0mn0m:
@connectbo good to know and thanks for the info. 

Do you want to make this part of your apps page PR?

connectbo:
I think the formatByte function you made would be helpful. Can start from there if you won't mind :)

n0mn0m:
> I think the formatByte function you made would be helpful. Can start from there if you won't mind :)

I don't mind at all 😄 go for it. Let me know if I can help out.

PR #56, opened 2021-04-22
Clean layout

connectbo:
This pr features:
- Refactored the old app page view into two separate page views: available/active
- All axios calls will be handled inside app/instance context provider or api.js file. Csrf token are added to axios global header in the api.js file, and axios global interceptor will handle 403 issue.
- Big components are extracted from view page as separate components
- Implemented two column listing on the available page view
- Refactored available app config slider, along with how we store and validate config spec from localstorage
- CPU/GPU slider has a step of 1, memory has a step of 0.25G. Notice I do convert all memory into bytes when it comes to slider, so in that way we don't need to worry about different unit. In the axios calls, memory will be formatted with appropriate unit.
- Change Apps to Workspaces on header

I know it might be too much for a single pr, so I try to focus on the cleanup and it should have a better overall code structure. Will have more housekeeping work in the future.

@n0mn0m I tested on my machine for a while and fix some bugs. If there are more issues occurs, please let me know.


.gitignore
```diff
@@ -18,3 +18,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+package-lock.json
```
n0mn0m, line 22: I think we want this committed to make sure we have the version we expect of each package in production and not just latest.

n0mn0m (commented):
Overall looks good. When do we want to collapse search into a tab instead of a top level route?

Do we still need the about view?

.gitignore
connectbo: You are right. There were some issues when I was trying to install the react-notifications and forgot to restore it back. Will do a rebase.

mbwatson: Yes, commit this @connectbo

mbwatson: Ahh I'm just a few seconds late!

src/components/card/card.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react'
 import styled from 'styled-components'
 
 const Wrapper = styled.div(({ theme }) => `
-  width: 100%;
+  width: 40vw;
```
mbwatson, line 5: I dont think we should change the card definition. This was to accommodate columns? Then better would be to put them in a grid or flex container when using cards, and let them fill 100% of their container. Otherwise cards are 40vw wide by default, which seems strange.

connectbo: Ahh right. I'll take it!

connectbo:
Just confirmed with Hannah that we are keeping the 'About' page and moving search to the tab under workspaces :) @n0mn0m

src/api/api.js
```diff
@@ -0,0 +1,21 @@
...
+    if (error.response.status === 403) {
+        window.location.href = window.location.origin + '/frontend'
+    }
+    return Promise.reject(error)
+})
+
+export const logoutHandler = (helxAppstoreUrl) => {
+    return axios.post(`${helxAppstoreUrl}/api/v1/users/logout/`)
+        .then(r =>
+            global.window && (global.window.location.href = `${helxAppstoreUrl}/accounts/logout/`))
```
n0mn0m, line 20: This should route to `/frontend`. The api POST will log them out, and redirecting them to `/frontend` will put them back on the login page. Eventually when we retire the old frontend this will become just `/` or we can setup a login page to redirect to at that time if the app needs a different root page.

src/components/app/app-card.js
```diff
@@ -0,0 +1,221 @@
...
+          <div className="specs">
+            <SpecContainer><SpecName>CPU</SpecName><SliderMinMaxContainer><SpecMinMax>Min: {minimum_resources.cpus} {minimum_resources.cpus > 1 ? 'Cores' : 'Core'}</SpecMinMax><Spec><b>{currentCpu}</b><Slider type="range" min={minimum_resources.cpus} max={maximum_resources.cpus} step="1" value={currentCpu} onChange={(e) => setCpu(e.target.value)} /></Spec><SpecMinMax>Max: {maximum_resources.cpus} {maximum_resources.cpus > 1 ? 'Cores' : 'Core'}</SpecMinMax></SliderMinMaxContainer></SpecContainer>
+            <SpecContainer><SpecName>GPU</SpecName>{minimum_resources.gpus === maximum_resources.gpus ? <SpecDefaultText>Default Setting: {minimum_resources.gpus} Core</SpecDefaultText> : <SliderMinMaxContainer><SpecMinMax>Min: {minimum_resources.gpus} {minimum_resources.gpus > 1 ? 'Cores' : 'Core'}</SpecMinMax><Spec><b>{currentGpu}</b><Slider type="range" min={minimum_resources.gpus} max={maximum_resources.gpus} step="1" value={currentGpu} onChange={(e) => setGpu(e.target.value)} /></Spec><SpecMinMax>Max: {maximum_resources.gpus} {maximum_resources.gpus > 1 ? 'Cores' : 'Core'}</SpecMinMax></SliderMinMaxContainer>}</SpecContainer>
```
n0mn0m, line 200: @connectbo a couple thoughts on this:

- Should the text stay left justified under `min` so the layout stays similar to the surrounding rows?
- Should we say `Locked Setting` instead of default? To me default is a recommended setup, but I can change it. In the scenario here we are trying to communicate that this setting cannot be changed/modified.
- The logic right now is based on counts, but I think that the logic should be that a setting is considered locked whenever min === max on that resource type. The min may be 0, but the max may be N, that isn't locked. The min could be 4000MB and the max may be 4000MB that's locked. Does that make sense?

image

src/components/app/app-card.js
```diff
@@ -0,0 +1,221 @@
...
+}
+
+const formatMemory = (mb, decimals = 2) => {
+  let bytes = toBytes(mb);
+  return formatBytes(bytes, decimals);
+}
+
+
+// each app config value from localstorage will be validated here
+const validateLocalstorageValue = (config, app_id, min, max) => {
```
n0mn0m, line 146: 👏

src/views/about.js
```diff
@@ -17,11 +17,11 @@ const HeLxIntro = styled.div`
   width: 80vw;
 `
 
-export const Home = () => {
+export const About = () => {
   return (
     <Container>
-      <Title>HeLx</Title>
-      <HomeContainer>
+      <Title>About HeLx</Title>
```
n0mn0m, line 23: Nit: May be worth asking Hannah if we want the title or just the text since they have navigated to the about page.

mbwatson (commented):
the apps page layout needs to be responsive. jump this layout back to a single column for smaller screen sizes.

Screenshot from 2021-04-23 09-06-03

src/components/app/app-card.js
connectbo: Those all makes sense to me. Let me poke around and would apply locking logic to cpu/memory as well.

src/views/about.js
connectbo: I think we would get rid of this page and have a menu link to the future Helx marketing website from meeting this morning.

n0mn0m (approved):

n0mn0m (changes requested):
Hey Bo I'm seeing this in chrome:

image

Do you have that on your end? Can pair if that would help.

n0mn0m:
@connectbo can you update the `docker-compose.appstore.yml` file volumes to:

```
    volumes:
      - $HOME/.kube/config:/home/appstore/.kube/config
      - $HOME/.minikube/:$HOME/.minikube/
      - ./build/frontend/index.html:/usr/src/inst-mgmt/appstore/frontend/templates/frontend/index.html
      - ./build/frontend/:/usr/src/inst-mgmt/appstore/frontend/static/frontend
```

For the latest image update in appstore? That will let you test your changes local again.

n0mn0m:
@connectbo let me know when you think this is ready and I can get an image into my bb namespace for final testing with live tycho interactions.

connectbo:
@n0mn0m @mbwatson  I update the grid container, should work for now. Let me know if there is any issues with the live tycho interactions. There are more responsive design todos and other tweaks this week, but I think that's too much for a single pr.

> Hey Bo I'm seeing this in chrome:
> 
> image
> 
> Do you have that on your end? Can pair if that would help.

n0mn0m:
@connectbo I built and put a custom image for this on https://helx.ahagerman.blackbalsam-cluster.edc.renci.org/frontend/ if you want to do any testing there.

connectbo:
@n0mn0m Thanks! I saw that bug.

n0mn0m:
@connectbo got a new image up with your latest changes (sorry missed a pull) https://helx.ahagerman.blackbalsam-cluster.edc.renci.org/frontend/react/workspaces. Fix looks good. 

image

Feel free to do some more testing, but I think this is good to go from my testing.

n0mn0m (approved):

connectbo:
@n0mn0m No problem. I will go ahead and merge. Did a bunch of testing, card component overall looks good. I notice the instance loading might take a while. Will add a loading indicator.

PR #57, opened 2021-04-27
update a running deployment

muralikarthikk:


docker-compose.appstore-tycho.yml
```diff
@@ -0,0 +1,19 @@
+version: "3"
```
n0mn0m, line 1: Will this be needed after the appstore update?

Makefile
```diff
@@ -75,6 +84,9 @@ clean:
 	rm -rf build
 	rm -rf node_modules
 
+#clean.tycho: remove tycho
+clean.tycho:
```
n0mn0m, line 88: Can you add this to `clean` so it's called automatically?

src/views/apps.js
```diff
@@ -224,6 +225,11 @@ export const Apps = () => {
   const [instances, setInstance] = useState();
   const [tab, setTab] = useState('Available');
   const [refresh, setRefresh] = useState(false);
+  const [modalOpen, setModalOpen] = useState(false);
```
n0mn0m, line 228: @connectbo how does this fit in with your ongoing changes?

Makefile
muralikarthikk: Yes, i initially thought of adding this to clean. But then again while reinstall, the changes to tycho would be lost and assumed it would be nice to separate these out.

I agree with adding this to clean as well. Will go ahead and make that change.

docker-compose.appstore-tycho.yml
muralikarthikk: This doesn't have to here necessarily but thought might be useful. You are referring to the docker-compose file, right?

n0mn0m: The compose file, and in general all the tycho components. Ideally this project just uses the latest appstore develop image. The reason it can't in this scenario is an ordering problem right? Because Tycho hasn't been updated in the appstore requirements file yet and the `workspace` pr there hasn't merged so this PR for `helx-ui` needs the added tycho custom here?

connectbo:
Thanks @muralikarthikk for bringing up this! I will check those out tonight and let you know my thoughts.

src/views/apps.js
connectbo: It might not be able to fit in directly, since the apps page now is not responsible for rendering specific elements, and each available / active will take care of all the logic. I will take a look and check those out, can make some rearrangement/adjustments if needed.

n0mn0m:
@muralikarthikk Bo's changes that were asked for by Hannah and Steve just merged. Probably worth checking those out. I have an image with them deployed at https://helx.ahagerman.blackbalsam-cluster.edc.renci.org/frontend/ if you want to interact with it there.

@connectbo any suggestion for Murali and the workspace update functionality he has been asked to implement by Steve?

n0mn0m:
@muralikarthikk with the compose using `appstore:develop-latest` do we need the tycho bits in here? Trying to keep as much of appstore and tychos code out of this repo and just have images facilitate any integration. Understand that sometimes it can cause ordering blocks in the workflow, but I think that we don't make those kind of changes as often.

connectbo:
There are lot of changes in the code, so @muralikarthikk if you pull the latest code from develop branch, you might have some conflicts. Basically the old apps page was renamed to workspaces, and available/active page view will have its own component. You should be able to find the logic in active.js file. Can you do some re-arrangements? Happy to hop on a call of that's helpful.

muralikarthikk:
Thanks for your work and suggestions on this guys.

@n0mn0m Yes, I totally understand. I will go ahead and delete this in my next commit.
@connectbo I will check the changes that you made and make necessary re-arrangements. I will definitely ask for help, if needed.

muralikarthikk:
@connectbo @n0mn0m - Refactored the code just like Bo suggested. Please review and let me know if these changes are good to go.

src/views/active.js
```diff
@@ -114,6 +137,59 @@ export const Active = () => {
...
+                                            name="cpu"
+                                            placeholder="1, 2...."
+                                        />
+                                        <Modal.FormLabel key="formKey">Memory in Mi</Modal.FormLabel>
+                                        <Modal.FormInput
+                                            ref={memory}
+                                            type="text"
+                                            name="memory"
+                                            placeholder="1000, 2500...."
+                                        />
```
connectbo, line 183: Do we need to validate the setting input value? My understanding is the updated setting should stay in the range of min~max resource provided in the app card. Otherwise we might receive an error from tycho.

muralikarthikk: Yes, we need to validate those inputs. I left those validations for now since I am not entirely sure on how to implement that. The re-naming part works as expected.

muralikarthikk:
@connectbo @n0mn0m - Added some validations on cpu and memory inputs based on resources on the app card.

PR #61, opened 2021-04-29
Feature/loading indicator

connectbo:
Add spinners for loading instances and apps.

### DIFF
--- a/src/views/active.js
+++ b/src/views/active.js
@@ -1,11 +1,12 @@
 import React, { Fragment, useEffect, useState } from 'react';
-import styled from 'styled-components'
+import styled, { useTheme } from 'styled-components'
 import { Button } from '../components/button'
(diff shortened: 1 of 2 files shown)


mbwatson:
@connectbo i see this adds a second loading indicator. consistency is key -- especially for things like this. try to use the same loading indicator across the entire application (or at least the same indicators for similarly functioning parts of the application).

connectbo:
@mbwatson Ok. The second loading spinner is only used inside the launch button, so whenever a launch request is sent, it will replace the 'Confirm' text inside the button. When loading apps/instances, it uses the original loading spinner. I can try to fit the original one inside the button.

n0mn0m (approved):
tested on https://helx.ahagerman.blackbalsam-cluster.edc.renci.org/frontend/ appreciate the update.

PR #66, opened 2021-05-06
Refactor/search

connectbo:
This pr:
- upgrade knowledge graphs to the dug search version
- move search as a subpage of workspaces, all search url path will started as `'/workspaces/search*`' and remove the top right search bar
- route `Available` and `Active` page to `/workspaces/available` and `/workspaces/active` to keep consistency
- refactored menu item component, so visiting subpages under workspaces will let 'Workspaces' menu item selected
- clean unused components and code


src/views/available.js
```diff
@@ -43,7 +44,7 @@ export const Available = () => {
...
...","minimum_resources":{"cpus":"1","gpus":0,"memory":"4000M"},"maximum_resources":{"cpus":"8","gpus":0,"memory":"64000M"}},"jupyter-ds":{"name":"Jupyter Data Science","app_id":"jupyter-ds","description":"Jupyter DataScience - A Jupyter notebook for exploring and visualizing data.","detail":"Includes R, Julia, and Python.","docs":"https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#jupyter-datascience-notebook","spec":"https://github.com/helxplatform/app-support-prototype/raw/master/dockstore-yaml-proposals/jupyter-ds/docker-compose.yaml","minimum_resources":{"cpus":"1","gpus":0,"memory":"4000M"},"maximum_resources":{"cpus":"8","gpus":0,"memory":"64000M"}},"octave":{"name":"Octave","app_id":"octave","description":"A scientific programming language largely compatible with MATLAB.","detail":"GNU Octave is a high-level language, primarily intended for numerical computations.","docs":"https://www.gnu.org/software/octave","spec":"https://github.com/helxplatform/app-support-prototype/raw/master/dockstore-yaml-proposals/octave/docker-compose.yaml","minimum_resources":{"cpus":"1","gpus":0,"memory":"4000M"},"maximum_resources":{"cpus":"8","gpus":0,"memory":"64000M"}}})
```
n0mn0m, line 47: Is this is here to prevent the list from being empty? If so a couple thoughts:

- If the server gives an error what's the chances it won't error when somebody tries to launch the app from this list?
- What is the plan for maintaining this list relative to the common app list?

n0mn0m (commented):
Overall looks good. Search should be a top level menu item and not a workspace. Thanks for the kg updates.

src/views/available.js
connectbo: No.. sorry these are just for error testing, I will remove them. I plan to add some notifications here to display errors, and the app listing components will says 'no apps available'.

n0mn0m (approved):
This is up at https://helx.ahagerman.blackbalsam-cluster.edc.renci.org/helx if you want to check it out, but it looks good to me.

PR #73, opened 2021-05-13
Feature/instance readable metrix

connectbo:
This pr:
- modify tableview on active page, so app name / creation time column can grow if needed
- format cpu/gpu listing, add conditional rendering
- post instance will carry memory in megabytes by default, so we can expect the same unit type when getting response from /instance, and then frontend will format it to the appropriate one when rendering.

### DIFF
--- a/src/views/active.js
+++ b/src/views/active.js
@@ -107,6 +107,7 @@ export const Active = () => {
         },
         {
             name: 'Memory',
+            // Note: Memory in the response does not contain unit, but it matches the one when we post them (MB by default).
             selector: (row) => { return formatMemory(row.memory + 'M') },
             sortable: true
         }, {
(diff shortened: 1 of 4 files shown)


src/views/active.js
```diff
@@ -90,21 +92,23 @@ export const Active = () => {
...
             name: 'GPU',
-            selector: 'gpus',
+            selector: (row) => { return row.gpus / 1000 + ' Core' + (row.gpus / 1000 > '1' ? 's' : '') },
             sortable: true
         },
         {
             name: 'Memory',
-            selector: 'memory',
+            // Note: Memory in the response does not contain unit, but it matches the one when we post them (MB by default).
+            selector: (row) => { return formatMemory(row.memory + 'M') },
```
n0mn0m, line 111: I think Steve will ask for this to be in GB to keep the number more human friendly and based on 617. We should probably also update Tycho to include the memory type on the instance, or guarantee to serve in MB, then the frontend can let the format function operate to get the most human friendly unit version.

n0mn0m: I read this wrong, and testing showed me that. Sorry. LGTM.

n0mn0m (approved):

PR #75, opened 2021-05-13
initialize timeout session

connectbo:


src/contexts/environment-context.js
```diff
@@ -30,20 +29,25 @@ export const EnvironmentProvider = ({ children }) => {
...
 
   return (
     <EnvironmentContext.Provider value={{
       helxSearchUrl: process.env.REACT_APP_HELX_SEARCH_URL,
-      helxAppstoreUrl: window.location.origin,
+      helxAppstoreUrl: helxAppstoreUrl,
       context: context,
-      helxAppstoreCsrfToken: csrfToken
+      helxAppstoreCsrfToken: csrfToken,
+      timeoutSeconds: 200
```
n0mn0m, line 50: Should this be the value from the user data minus 100 or a percentage of that timeout instead of a hard coded value? Could be annoying it a timeout is set to 20 minutes, but we pop this up more often.

connectbo: Right, timeout value from /user is loaded and will be consumed. I set this for local testing purpose, sorry for the confusion!

PR #102, opened 2021-06-10
Bugfix/pre search loading spinner

mbwatson:
this PR
- solves the rendering of the loading indicator before search
- removes the results selection logic (bc it isn't used)
- adds a clear button to the search input

@connectbo i'd appreciate a second set of eyes on this solution to the first bullet above.


connectbo (approved):

connectbo:
Looking good to me. Thanks for fixing that, merging.

PR #114, opened 2021-06-14
Enhancement/search interface tweaks

mbwatson:
this improves the search interface in a couple small ways.

- prevent the search input from stretching across a very large screen by setting a max width of 1080px.
- center (vertically) the search input until search results are returned, then animate input to top of page.


connectbo (approved):
Thanks. Search looks much nicer! Merging

PR #118, opened 2021-06-24
Feature/layout config

mbwatson:
this adds a button group to toggle search result card layout, allowing the end user to render search results as a grid (the only option currently) or as a list.


connectbo (approved):

connectbo:
looking good, merging!

PR #123, opened 2021-06-30
Feature/disable search

connectbo:
This pr
- update the sample .env file, adding two new boolean variables to control the rendering of search and workspace modules
- add conditional rendering in menu items and views
- refactor homepage route logic: If both search and workspace are enabled, '/' will be routed to search page by default. If search is disabled, hitting '/' will be routed to workspace page. If both are disabled, route baseurl to support page.

Let me know if this is a good approach.


src/contexts/environment-context.js
```diff
@@ -37,6 +37,8 @@ export const EnvironmentProvider = ({ children }) => {
     <EnvironmentContext.Provider value={{
       helxSearchUrl: process.env.REACT_APP_HELX_SEARCH_URL,
       helxAppstoreUrl: window.location.origin,
+      enableSearch: process.env.REACT_APP_ENABLE_SEMANTIC_SEARCH,
```
mbwatson, line 40: this is so minor, but having a verb as the first part of this variable makes me think that it might _do_ something, that it might be a function. i think maybe something like `searchEnabled` indicating that this is a thing that has been defined already, since it's in past tense. does that make sense?

mbwatson (commented):
i think this all looks great. quick and easy solution. i even think we can re-use some of this if/when we change the approach for this feature.

src/contexts/environment-context.js
connectbo: Yep, I think that sounds better! Will change those.

connectbo:
Renamed env variables, merging.

PR #131, opened 2021-07-07
Feature/lazy load search

mbwatson:
The changes in this PR accomplish two main goals:

1. **Organization:** The search component tree was getting a little cluttered, so i've grouped search components with multiple files into directories

2. **Search Lazy Loading:** There are three search requests to get (a) the initial results, (b) the studies, and (c) the knowledge graphs (KGs). Searching for all of these entities at once fires of about 80 requests after pressing the _Search_ button, which causes some borderline noticeable performance hits. On a slower connection, this is problematic. Here, this is solved by moving the logic for the two requests so they only fire off when the studies or KGs tab is selected (hence "lazy" -- we get the things only when you need the things). This brings the initial search request count down to _four_, significantly increasing load time/performance/user experience.

However, there is one caveat: we don't know how many knowledge graphs (perhaps none!) are present before selecting the KGs tab. This UX trade-off is well worth it IMO, and we can solve the not-knowing-how-many-KGs-there-are problem later.

@connectbo What do you think?


connectbo (approved):
This makes search feels much smoother than before. Looking good!

PR #132, opened 2021-07-13
remove invalid results from display & log details to console

mbwatson:
this adds a snippet to the search context to
- separate results into two buckets &mdash; valid and invalid
- only render the valid bucket
- dump the invalid bucket to the browser console


connectbo (approved):
This is really a fast and clean way to filter invalid results. Thanks.

PR #137, opened 2021-07-19
Refactor/configurable modules

connectbo:
This pr
- remove env vars from Dockerfile (for now), so the build process won't pick up preset environment variables
- add a simple loading spinner, which would show up when a request to /context endpoint is in process. Since we are sending out request to that endpoint every time on page refresh for now, which means you will see the spinner every time you click 'refresh'. Is that problematic? @mbwatson How do you think of using sessionStorage to store context data (brand, env vars)? 
- refactored the app top level architecture, let me know if that is a good approach or not.


connectbo:
I got this deployed on my blackblasm [namespace](https://helx.zbo.blackbalsam-cluster.edc.renci.org) with search off and workspace on. Feel free to take a look.

connectbo:
@mbwatson I experimented with that approach I mentioned yesterday a little bit. Sessionstorage is now used to store context data after initial render, which means we don't need to send out request to /context every time on page refresh. Let me know if that makes sense.

mbwatson (approved):

PR #141, opened 2021-08-04
Feature/connect button

connectbo:
This pr
- click on connect button on instance page will take you to a splash screen and trigger the polling service
- brings back the splash screen we used to have and modified its layout, so it wont display the top navbar
- check and redirect if the url is ready

I also get it deployed on my [namespace](https://helx.zbo.blackbalsam-cluster.edc.renci.org/) so you can review and try it :)


mbwatson (approved):
works as expected!

PR #143, opened 2021-08-10
Feature/env startup

cschreep:
Re-opening this pull request because github was not picking up all the commits properly


connectbo:
@cschreep I updated the ui components and context provider so each env var can be picked up and render correctly. I also replace the start script delimiter '/' with '!' as we will have slash in the url value. Using slash as delimiter will cause collision. Let me know if that's ok.

cschreep:
@connectbo This looks good. The only other things I think we should do is 1. remove the code that has been commented out, and 2. In the sed script I think a more common approach for handling arbitrary values is to stick with `/` as a delimiter, and escape it in the env var definition (e.g. `http:\/\/helx.renci.org`)

connectbo:
Ah yes, that makes sense. Just pushed the changes, gonna merge if it looks good.

cschreep:
Have you built a docker image with these changes?

connectbo:
Yes, I built an image and it could load env vars at runtime. One thing I can think of is, do we need to update the appstore chart? I know we put env vars in extraEnv in the past and let `/context` to pass along these vars. For now, we can just expose those env vars and ui should be able to pick those up.

cschreep:
Yeah, we will have to update it. We have a config map in appstore that defines the available env vars, so we can just update that and make sure everything is passed in appropriately, but this is good to merge on its own now

connectbo (approved):

PR #150, opened 2021-08-19
show study variables in result modal

mbwatson:
merging this PR will change how the studies lists are rendered in the result modal.

a collapsible list of variables for each study associated to the result will be shown. note that this adds this for the modal only, to prevent cluttering up the smaller result cards.


cschreep (approved):
Looks legit

connectbo (approved):
This approach looks good to me.

PR #156, opened 2021-09-15
Feature/load instance count limit

connectbo:
This pr includes the following ui enhancements:
- loaded max app instance count from `/apps` and disabled launch button when it reaches the limit.
- ui will be redirected to available tab when all apps are closed

Got it deployed on my [namespace](https://helx.zbo.blackbalsam-cluster.edc.renci.org/), please take a look. Let me know if you have any thoughts or ideas!


mbwatson:
a user may not know why the Launch button is disabled, so let's do a tooltip for the disabled launch button that says something along the lines of "the maximum number of instances is already running"

mbwatson (approved):
:+1:

PR #157, opened 2021-10-06
add argus brand logo and fix issue

connectbo:
This pr should let ui
- supported argus branding
- rendered helx logo if no specific brand is defined
- fixed the logout mobile button issue by passing in the appstore url

image


mbwatson (approved):
looks good

PR #165, opened 2021-11-03
Enhancement/respond to other status code

connectbo:
This pr includes the following changes
- UI will react to status code `403` when the instance is not created by the one requesting the delete, which won't normally happen 
- UI will react to status code `404` when the instance no longer exists

image

Got it deployed on my namespace if you want to take a look: https://helx-dev-zbo.apps.renci.org/

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -13,7 +13,7 @@ server {
     }
 
     location / {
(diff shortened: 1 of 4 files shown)


waTeim (commented):
looks reasonable

mbwatson (approved):
:+1:

PR #166, opened 2021-11-11
Feature/tranql

frostyfan109:
This PR adds TranQL functionality as a tab in the DUG result modal.
- Adds an interactive iframe of the embedded version of TranQL using a query derived from the dug result.
- Adds links to open in full TranQL version and to open Robokop
- Adds the environment variable `REACT_APP_TRANQL_ENABLED`, which is `false` by default.

It also:
- Refactors result modal tabs into their own directory and removes extraneous imports from the result modal.
- Fixes `Link`/`ExternalLink` so that custom props may be passed to an external link.
- Adds the use-debounce dependency.

Note that basic analytics tracking is enabled on the TranQL tab (i.e. it will track that a user is using the tab and when a user opens the links to full TranQL/Robokop), but it will not track in-depth analytics related to the usage of TranQL itself (as it is in an iframe and the work/tech required to add something as such probably isn't worthwhile).


mbwatson (approved):
looks good to me!

PR #168, opened 2021-11-17
Conditional rendering of support page components

connectbo:
This pr removes the breadcrumb and allows users to hide 'Community Support' or 'Documentation' section on the `/support` page. (i.e. for bdc, we don't have Community Support/discourse ).

This requires a new env variable `hidden_support_section` to be added to the ui-chart.
related PR: https://github.com/helxplatform/ui-chart/pull/9

### DIFF
--- a/bin/populate_env
+++ b/bin/populate_env
@@ -12,6 +12,7 @@ search_url="${REACT_APP_HELX_SEARCH_URL}"
 brand_name="${REACT_APP_UI_BRAND_NAME}"
 tranql_enabled="${REACT_APP_TRANQL_ENABLED:-false}"
 analytics="${REACT_APP_ANALYTICS:-0c39f8410fd548bfa1976957d0248289}"
+hidden_support_section=${REACT_APP_HIDDEN_SUPPORT_SECTION}
 
 
 template='{
@@ -28,7 +29,8 @@ template='{
       "mixpanel_token": "%ANALYTICS%",
       "ga_property": ""
       }
-    }
+    },
+    "hidden_support_section": %HIDDEN_SUPPORT_SECTION%
 }'
 
 echo $template | sed \
@@ -36,6 +38,7 @@ echo $template | sed \
   -e "s/%SEARCH_ENABLED%/$search_enabled/" \
   -e "s/%SEARCH_URL%/$search_url/" \
   -e "s/%BRAND%/$brand_name/" \
+  -e "s/%HIDDEN_SUPPORT_SECTION%/$hidden_support_section/" \
   -e "s/%ANALYTICS%/$analytics/" \
   -e "s/%TRANQL_ENABLED%/$tranql_enabled/" \
   > $1
--- a/bin/populate_env
+++ b/bin/populate_env
@@ -12,7 +12,7 @@ search_url="${REACT_APP_HELX_SEARCH_URL}"
 brand_name="${REACT_APP_UI_BRAND_NAME}"
 tranql_enabled="${REACT_APP_TRANQL_ENABLED:-false}"
 analytics="${REACT_APP_ANALYTICS:-0c39f8410fd548bfa1976957d0248289}"
-hidden_support_sections=${REACT_APP_HIDDEN_SUPPORT_SECTIONS}
+hidden_support_sections="${REACT_APP_HIDDEN_SUPPORT_SECTIONS}"
 
 
 template='{
@@ -30,7 +30,7 @@ template='{
       "ga_property": ""
       }
     },
-    "hidden_support_sections": %HIDDEN_SUPPORT_SECTIONS%
+    "hidden_support_sections": "%HIDDEN_SUPPORT_SECTIONS%"
 }'
 
 echo $template | sed \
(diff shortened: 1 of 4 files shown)


mbwatson (changes requested):
your solution makes sense to me.

this feels more appropriate than the obvious alternative, which is to have the opposite variable, something like `visible_support_sections`. however, that approach would require updating all deployments' configurations if/when a new support section is added so it becomes visible. so i think this is good. 

i do have one teeny note. can you pluralize the `hidden_support_section` variable so it's clear on both ends that it is an array value?

connectbo:
> your solution makes sense to me.
> 
> this feels more appropriate than the obvious alternative, which is to have the opposite variable, something like `visible_support_sections`. however, that approach would require updating all deployments' configurations if/when a new support section is added so it becomes visible. so i think this is good.
> 
> i do have one teeny note. can you pluralize the `hidden_support_section` variable so it's clear on both ends that it is an array value?

Yeah, you are right, that will explain more. I will make changes on both ends.

src/views/support.js
```diff
@@ -1,30 +1,36 @@
 import { Fragment } from 'react'
 import { Typography } from 'antd'
-import { Breadcrumbs } from '../components/layout'
+import { useEnvironment } from '../contexts'
 
 const { Title } = Typography
 
+const CommunitySupport = () =>
+  <Fragment>
+    <Title level={1}>Community Support</Title>
```
mbwatson, line 9: as secondary sections, these should have level={2}

bin/populate_env
```diff
@@ -30,15 +30,15 @@ template='{
       "ga_property": ""
       }
     },
-    "hidden_support_section": %HIDDEN_SUPPORT_SECTION%
+    "hidden_support_sections": %HIDDEN_SUPPORT_SECTIONS%
```
cschreep, line 33: Pursuant to my comment on the helm chart PR, I think we should expect this env var to be a comma-separated string, and we should parse it into an array in this script

cschreep (changes requested):

bin/populate_env
cschreep: (related comment: https://github.com/helxplatform/ui-chart/pull/9#discussion_r752586364)

connectbo: Ahh I see, we need to split the string into an array in bash, does the following syntax look good? 
`hidden_support_sections=(${REACT_APP_HIDDEN_SUPPORT_SECTIONS//,/ })`
First we replace all comma with space and the surrounding parenthesis will covert that into array?

cschreep: I think there's an extra step you need, since we have to go from `"foo,bar"` to `["foo","bar"]` so we need to wrap each individual string in double-quotes to make it valid JSON

connectbo: I'm not super familiar with the syntax, would the following work?

```
for i in "${!hidden_support_sections[@]}"
do 
    hidden_support_sections[$i]="${hidden_support_sections[$i]}"
done
```

cschreep: Maybe, but I'm also thinking piecing it together in bash might not be the ideal approach. Now that I think about it a little more, it might be easier to leave it as a comma-separate string in the env.json, and convert it into an array in the Javascript code when you load the environment params into the window. Would that be easier to implement?

connectbo: yeah, actually I was thinking about that too. This is a little bit unnecessary. UI can get the string can just split it there.

src/contexts/environment-context.js
```diff
@@ -64,6 +64,9 @@ export const EnvironmentProvider = ({ children }) => {
       url: `${relativeHost}/static/frontend/env.json`
     })
     let context = response.data;
+
+    // split the comma-separated string which tells ui the support section to hide
+    context.hidden_support_sections = context.hidden_support_sections.split(',')
```
cschreep, line 69: This is definitely the easier approach

bin/populate_env
```diff
@@ -12,6 +12,7 @@ search_url="${REACT_APP_HELX_SEARCH_URL}"
 brand_name="${REACT_APP_UI_BRAND_NAME}"
 tranql_enabled="${REACT_APP_TRANQL_ENABLED:-false}"
 analytics="${REACT_APP_ANALYTICS:-0c39f8410fd548bfa1976957d0248289}"
+hidden_support_sections="${REACT_APP_HIDDEN_SUPPORT_SECTIONS}"
```
cschreep, line 15: Is there a good default value to set for this?

cschreep (approved):

bin/populate_env
connectbo: hmm, I think the default is none? By default, all sections on /support page should be there. The initial goal for this pr is to hide the community section on bdc.

src/contexts/environment-context.js
connectbo: yep, I will update the ui-chart pr shortly.

bin/populate_env
cschreep: ok, as long as the UI is tested with an empty string here and works as expected I say 👍🏻

connectbo: yep, an empty string will just leave all sections there. Going to merge. Thanks for the help Carl!

PR #169, opened 2021-11-19
Feature/analytics update

frostyfan109:
Adds a couple additions to the analytics tracking for all events:
- Every event will have the brand/deployment it's coming from attached.
- Every event will now report the user id or `REMOTE_USER` as long as the user is signed in and `workspaces_enabled` is set to true.

Also upgrades to `helx-analytics@1.0.14` which adds support to an Event's customParameter field for Function and Promise values.


mbwatson (approved):
:+1:

PR #171, opened 2021-12-09
Bug/delete app cleanup

connectbo:
This pr is intended to clean up messages in the side drawer and stop polling when an instance is stopped.

related: https://github.com/helxplatform/development/issues/839

### DIFF
--- a/src/contexts/instance-context.js
+++ b/src/contexts/instance-context.js
@@ -10,13 +10,21 @@ export const InstanceProvider = ({ children }) => {
     const { helxAppstoreUrl } = useEnvironment();
     const { addActivity, updateActivity } = useActivity();
     const [openedTabs, setTabs] = useState([]);
+    const [pollingTimerIDs, setpollingTimerIDs] = useState([])
+
+    const stopPolling = (sid) => {
+        // stop polling when instance is deleted
+        clearTimeout(pollingTimerIDs[sid])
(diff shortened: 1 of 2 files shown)


connectbo:
@mbwatson I think this pr is ready for review. When an instance is not up and ready but you want to delete it, its polling function will break and the 'xxx is launching' message will be removed from the side drawer. I got this deployed in my [namespace](https://helx-dev-zbo.apps.renci.org/helx) if you want to click around. Feel free to share your thoughts.

src/contexts/instance-context.js
```diff
@@ -10,13 +10,21 @@ export const InstanceProvider = ({ children }) => {
     const { helxAppstoreUrl } = useEnvironment();
     const { addActivity, updateActivity } = useActivity();
     const [openedTabs, setTabs] = useState([]);
+    const [pollingTimerIDs, setpollingTimerIDs] = useState([])
```
mbwatson, line 13: teeny thing, but i think you wanted a capital "P" in setpollingTimerIDs

mbwatson (approved):
it works! and this is a nice solution.

src/contexts/instance-context.js
connectbo: Nice catch on this! Right, definitely should be a capital 'P' there, will change that. Thanks.

PR #172, opened 2021-12-10
Tranql updates

frostyfan109:
This PR contains improvements to the HeLx-UI/TranQL integration.
- Moves "Open in Robokop" link to its own "Robokop" tab.
- Move "Open in TranQL" link to floating above the TranQL query codemirror component.
- Adds a loading screen to the iframe to avoid large empty whitespace block.
- Fixes sizing of iframe so that the iframe and TranQL components will properly resize when the page is stretched and then shrinked again.
  - Remove `react-sizeme` dependency and size iframe with fixed height and 100% width (css).
- Add new `?use_last_view` query param to TranQL iframe so that the last used view will be the default view on load.
- Add `LIMIT 100` clause to TranQL queries to ensure fast load times.


connectbo (commented):
Code looks good to me, the only concern I have is the floating 'View in tranql' link may overlap the 'Run' button a little bit.

image

connectbo (approved):

mbwatson (approved):
:+1:

PR #174, opened 2022-01-21
Fixes #173

### LINKED ISSUE #173: Bug with filter: drop-shadow on antd cards in Safari
Causes a large partially-opaque "banner" to appear at the bottom of the page which obscures the content.

image

Changing drop-shadow to red:

image

frostyfan109:
Adds a `display: none` rule to the ::after pseudo-element which was causing issues, since it didn't seem to be performing any function in the first place.

### DIFF
--- a/src/components/search/result-modal/tabs/tranql/index.js
+++ b/src/components/search/result-modal/tabs/tranql/index.js
@@ -22,6 +22,8 @@ export const TranQLTab = ({ result, graphs }) => {
   const { query } = useHelxSearch()
   const tranqlUrl = context.tranql_url
 
+  const robokopUrl = "https://robokop.renci.org"
+
   const trackLink = (name, url) => {
     analytics.trackEvent({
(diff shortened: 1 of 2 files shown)


mbwatson (changes requested):
i think that instead of using display: none for that pseudoelement, we should just remove it from the code. along with the same reference on lines 54-56. what do you think about that?

frostyfan109:
Updated the fix addressing #173
- Change pseudo-element's `linear-gradient` so that it works properly with Safari (only has partial support). 
- Change pseudo-element's `position:fixed` relative to `.result-card.ant-card` to be explicitly  `position:absolute` relative to `.result-card.ant-card` for browser compatibility.

Fix padding/text positioning issue with `.ant-card-body` where text appeared to have extra top padding due to an inline display rule applied by `.ant-space`.

> Before
image

> After
image

mbwatson (approved):
👍

PR #175, opened 2022-02-10
Move ROBOKOP into a "link" tab

frostyfan109:
- Adds a link container inside the result modal menu alongside the tab container.
- Moves ROBOKOP into this link container.


connectbo (approved):
LGTM

PR #176, opened 2022-02-16
Fix mobile card sizing and fix ROBOKOP link capitalization.

frostyfan109:
- Fixes card sizing on mobile devices to prevent unwanted horizontal overflow.
  - Uses a sm breakpoint (mobile devices) to manually adjust `grid-template-columns` to 250px min instead of 400px min (since mobile devices can go down to around 300px wide and there is a good amount of horizontal padding that needs to be considered).
  - This generally doesn't affect how larger mobile devices and desktop devices size cards, but fixes sizing on small (<450px) devices.
- Capitalizes the ROBOKOP link (forgot to in previous PR).


connectbo (approved):
looks good to me

src/components/search/results/results.css
```diff
@@ -24,6 +24,13 @@
   grid-template-columns: 1fr;
 }
 .results .results-list.grid {
+  /* Phone screens can get as small as 300px wide, so it's necessary to adjust the
+   * minimum size of cards to below this so that they don't overflow the screen 
+   * (and have to adjust for the fact that about 50-75px of screen width is padding). */
+  grid-template-columns: repeat(auto-fit, minmax(250px, 1fr));
```
mbwatson, line 30: good catch

src/components/search/results/results.js
```diff
@@ -81,7 +86,7 @@ export const SearchResults = () => {
           <div className="results">
             { results.length >= 1 && MemoizedResultsHeader }
 
-            <div className={ layout === GRID ? 'results-list grid' : 'results-list list' }>
+            <div className={gridClass}>
```
mbwatson, line 89: :+1:

mbwatson (approved):
looks good!

PR #179, opened 2022-02-18
Introduce variable search

gingin77:
@mbwatson, @connectbo and I paired on this

We were looking at Bo's ant-v implementation had some reasons to take a step back. We reverted the name `ConceptResults` back to `Result` per the discussion with Steve C and other DUG developers


gingin77:
Closing this for now... may need to rebase later, so will create new PR when needed

PR #180, opened 2022-02-24
Adds a variety of analytics updates

frostyfan109:
- Adds deployment namespace tracking to all events (to more easily differentiate between prod and dev deployments)
- Adds analytics tracking to workspaces (in anticipation of analytics possibly being added to eduhelx and other workspace-enabled deployments)
  - Also adds logout tracking
- Adds route-tracking events to all paths, i.e. a routing event is fired when upon page redirects within the app
- Refactors the analytics context to cut down on the only tangentially-related code blocks introduced by tracking analytics events within components.


connectbo (approved):
👍

mbwatson (approved):
:+1:

PR #183, opened 2022-04-01
Studies histogram gh

gingin77:
Hi @mbwatson I updated the name of the function called by the pushpin to `toggleVariablesByStudy` and handled variable bar selection accordingly.

Also, I muted the color. Aquamarine felt a little too extra.

There's still an issue where bars selected by study get lost when a user zooms in, but I think we are probably in an ok place to meet with Hannah now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ yarn-error.log*
 # IDE files
 .idea/
 public/env.json
+misc/
--- a/package-lock.json
+++ b/package-lock.json
@@ -12855,6 +12855,14 @@
       "resolved": "https://registry.npmjs.org/react-ga/-/react-ga-3.3.0.tgz",
       "integrity": "sha512-o8RScHj6Lb8cwy3GMrVH6NJvL+y0zpJvKtc0+wmH7Bt23rszJmnqEQxRbyrqUzk9DTJIHoP42bfO5rswC9SWBQ=="
     },
+    "react-infinite-scroll-component": {
+      "version": "6.1.0",
+      "resolved": "https://registry.npmjs.org/react-infinite-scroll-component/-/react-infinite-scroll-component-6.1.0.tgz",
+      "integrity": "sha512-SQu5nCqy8DxQWpnUVLx7V7b7LcA37aM7tvoWjTLZp1dk6EJibM5/4EJKzOnl07/BsM1Y40sKLuqjCwwH/xV0TQ==",
+      "requires": {
+        "throttle-debounce": "^2.1.0"
+      }
+    },
     "react-is": {
       "version": "16.13.1",
       "resolved": "https://registry.npmjs.org/react-is/-/react-is-16.13.1.tgz",
@@ -15051,6 +15059,11 @@
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/throat/-/throat-5.0.0.tgz",
       "integrity": "sha512-fcwX4mndzpLQKBS1DVYhGAcYaYt7vsHNIvQV+WXMvnow5cgjPphq5CaayLaGsjRdSCKZFNGt7/GYAuXaNOiYCA=="
+    },
+    "throttle-debounce": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/throttle-debounce/-/throttle-debounce-2.3.0.tgz",
+      "integrity": "sha512-H7oLPV0P7+jgvrk+6mwwwBDmxTaxnu9HMXmloNLXwnNO0ZxZ31Orah2n8lU1eMPvsaowP2CX+USCgyovXfdOFQ=="
     },
     "through": {
       "version": "2.3.8",
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "helx-analytics": "^1.0.14",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "react-infinite-scroll-component": "^6.1.0",
     "react-scripts": "4.0.3",
     "styled-components": "^5.3.0",
     "timeago-react": "^3.0.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "helx-ui",
-  "version": "2.0.0-dev.13",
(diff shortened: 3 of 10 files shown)


gingin77:
@mbwatson - oh great and thanks! I have a meeting with Hannah and Kira tomorrow morning that will likely lead to some tweaks and possibly new bugs revealed.

frostyfan109:
@gingin77 This is really awesome. Excited to see this added in the near future.

FYI I will handle the merging into develop since #186 made some major changes to how layout is handled.

Notes:
- I'm guessing the icon is just a placeholder, though I would suggest `AreaChartOutlined` or `BarChartOutlined`.
- There should be a bit more of a margin (thinking 16px) between the vertically-spaced components to stay consistent with other parts of the site.
- It may be nice to have the pinned studies appear at the top of the list so that you can easily tell what is pinned/unpin them.
- The results header is a little wordy now, the extra study info may be more appropriate in another container elsewhere, not really sure.
- The button container may make more sense below the graph. Then the "variables according to dug" text could be made into a header i.e. using `Typography.Title`. Something like this:
image


Bugs I found:
- There is some strange hanging occurring when a search is executed. Memoizing `ConceptsList` helped, but there seems to be more to it than just that.
- There needs to be some sort of enforced zoom limit. If you keep zooming past the single-variable level (because of course some people like me will 🙂) it can crash the page.

frostyfan109:
@gingin77

Thanks for taking a look at my review.

Just to be clear, I haven't really reviewed any of the code. I looked around a bit but nothing critical. I'd be happy to do an actual code review in the future before the merge to develop if that would be helpful.

One thing I'm noticing this time around is that `variable-results.js` is bit of a behemoth. I'm not sure how much of this can actually be separated out, though ideally the entire layout would not be encapsulated within a single file/hook.
- At the most basic level, I'd separate the histogram and variable/study lists into two standalone components.

> I'm guessing the icon is just a placeholder, though I would suggest AreaChartOutlined or BarChartOutlined.

I was referring the ellipsis icon used in the grid selection component:

image

> There should be a bit more of a margin (thinking 16px) between the vertically-spaced components to stay consistent with other parts of the site.

Vertical spacing could be added via hard-coded margins or `Space`. For reference:
- `<Space size="small">` corresponds to a 8px margin, and this is the default when `size` isn't given.
- `<Space size="middle">` is 16px.
- `<Space size="large">` is 24px.

You could also use `24px` margins if you think it looks better. I just wouldn't use anything less than `16px`. Regardless, I think using `<Space>` is probably best here. There are a number of places in the code that don't use Space, and instead use hard-coded margins or replicate the behavior of Space, but this is just because there are drawbacks to how Space works that mean it sometimes can't be used well.

> I'm remembering that there is some file clean-up/reorganizing that we should do.

The project is definitely in need of some cleanup in general. This hasn't been addressed as far as I know; Matt hasn't been working much on the UI, and I haven't gotten around to it either.

> The results header is a little wordy now, the extra study info may be more appropriate in another container elsewhere, not really sure.

I didn't really explain my thought process much here. I think the info is useful to have, it's just that when you look at the header now, there are 4 different metrics a user has to take in from a single "sentence" (concept count, study count, variable count, page count), and it's just a bit too much for me. And to make it even worse, #191 changed the header to (x of y pages) so there'd be __5__ things now.

> There is some strange hanging occurring when a search is executed. Memoizing ConceptsList helped, but there seems to be more to it than just that.

I think I said this because beforehand React was memoizing `ConceptsList`. You changed some stuff around in `results.js` and it isn't anymore. So whenever any of the *many* pieces of state change that `SearchResults` uses from the search context, it induces a re-render of `ConceptsList`.

I personally haven't ran into performance problems with the studies layout, but breaking it down into more components would probably help improve performance to some extent.

gingin77:
- [ ] Summary sentence - 4 different metrics a user has to take in from a single "sentence" (concept count, study count, variable count, page count)
- [ ] Vertical spacing 
- [ ] Separate the histogram and variable/study lists into two standalone components
- [ ] Bring back "View Last Zoom"
- [ ] Start conversation with Hannah, Kira, Alex regarding the icon for the Toggle - Concepts vs Variables - should we use Histogram Icon?

gingin77:
I think there's something wrong related to how I'm using `setVariableStudyResults` vs `setVariableResults` - will return to this next chance I can keep debugging/making fixes

gingin77:
Re commit [b66ad29](https://github.com/helxplatform/helx-ui/pull/183/commits/b66ad29ce0a09db4951e9d593c74d252735f9343), use `Title` instead of `Text`

frostyfan109:
@gingin77 These are some ui features I'd like to see after messing around with the view some more:

- [ ] Title/header improvements (beyond just "Variables according to DUG score".
  - I think an addition like `Zoomed into variables with scores in 90-94th percentile` or something along those lines may be a good approach to help the user keep track of where in the overall histogram they are zoomed into.
- [ ] Pinned studies should appear at the top of the list
  - Sorry, this ended up getting lost somewhere down the line, and we didn't end up discussing this. I think intuitively, a user expects a pinned UI component to appear at the top of a list, but this isn't the current behavior. E.g. in the screenshot below, I have the study "NIDA-CTO0010" pinned, but it doesn't appear at the top of the list.
image


- [ ] Margins need to be cleaned up on the studies list
  - Specifically the left borders next to each variable under a study clash with the overall border of the study container, so these need to be properly spaced out.
  - There's also too much space between the study name/pushpin header and the variables list container.
image

- [ ] The left border next to each variable under a study is very difficult to see due to the gray background.
  - Either it needs to be darker, or you could remove it and do something like I've done with the CDEs UI instead (see screenshot). The left handle turns blue when hovered, and collapses the interface when you click on it.
image


- [ ] There should be some sort of interface from left clicking a variable in the histogram or studies list. Otherwise, the interactivity of the interface will be pretty limited overall. It may be worthwhile to have a discussion with some more people about what this interface would do and what it would look like.
  - One idea I have is that it opens an interface to select concepts associated with the variable/study, and then maybe you can select a concept and view it, though maybe this isn't the direction we want to go in for the variables layout.
  - Maybe it suggests other searches/concepts associated with the variable/study that you can redirect the search to.
  - Etc.

PR #186, opened 2022-05-24
Expanded result layout

frostyfan109:
Adds a new "expanded" browsing layout for search results that dedicates more screen real estate to viewing info about a specific result.
- New layout (will be the default layout going forward).
- Removes "list" layout, so just new layout/grid layout are left.
- User's most recently selected layout is stored in local storage and automatically used to on page load. 

Misc. unrelated changes:
- Change TranQL/Robokop icons to be more relevant.
- Fix copy link notification.
- Remove excessive footer margin.

Note: draft until unit tests are passing


frostyfan109:
Tests failing due to Jenkins availability, but run locally.

mbwatson (approved):
this is the best improvement to the dug ui in a long time. thanks so much for working through this.

i wonder if instead of "No result selected" we put something instructive there, like "Please select a concept from the drawer to view it in detail. OR maybe the first one is selected by default?

i'll go ahead and approve, though, as these details can be addressed later.

PR #191, opened 2022-06-09
Concept-type filtering, infinite scrolling, eslint cleanup, & unrelated bug fixes

frostyfan109:
### Infinite scrolling
The grid and expanded layouts both now use infinite scrolling instead of explicit pagination to load more results.

A "scroll to top" FAB will also appear at the bottom right for easily scrolling back to the top of the page after looking through results.

A couple notes:
- Pagination is still supported internally, but is no longer supported through the query string, i.e. `?p=5` will simply be ignored and page 1 will be loaded.
- `concepts` is still supported in the search context for accessing the current page, but `conceptPages: { [page: number]: concept[] }` can be accessed to get any of the loaded pages.
- Feedback would be helpful here: do we want to still support loading results through the query string? I.e. would it be useful to a user to be able to go to `/search?q=asthma&p=4`, and the infinite scrolling will start at page 4 instead of page 1 (without any means to access the results of prior pages, perhaps with indication that this is the case).

### Type filtering
Introduces a select menu for client-side filtering based on concept type as outlined [here](https://github.com/helxplatform/dug/issues/231).

For now, this will remain a client-side thing, but with the addition of infinite scrolling, phasing out explicit pagination, it should be in a good place to change to an api parameter sometime in the future.

The obvious limitations of the current client-side implementation are:
- Say someone wants to view just `biological processes` for the search `heart`. There is only 1 biological process out of the 167 concepts this search returns. A user would have to scroll through all 9 pages, and would only find one. Ideally, a search for biological processes under hearts would be independently paginated from a general search for hearts, and thus this search would only yield 1 concept to begin with.
- Loading new pages has been disabled when filtering by a concept type. This can easily be fixed by a server-side implementation. The simple reason is that when filtering by an uncommon type, one may go a lot of pages in a row without ever getting a new result. This sort of ui behavior is really unintuitive for a user. There is also an infinite scroll-related reason that is a bit more convoluted.

### Linting fixes
Went through and fixed most of the linting warnings that had been building up for a while.

### Unrelated bug fixes/changes
- The "search focus" shortcut now will trigger off of both forward slash and backslash. This is since forward slash is a more [familiar and accepted accepted shortcut](https://superuser.com/questions/1461734/is-there-a-common-browser-shortcut-for-jumping-to-a-website-s-search-bar) for this sort of behavior, but backslash has been left as well in order to avoid any confusion.
- The "search focus" shortcut now simply focuses the input and scrolls it into view, rather than also selecting the current text.
- The "search focus" shortcut has been fixed so that a user can enter backslash/forward slash in the input (before these inputs were intercepted by the shortcut and would not allow the user to input these characters)
- An old version of ant design's CSS still being used in the project (v4.6.6), despite being on ant design v4.20.5. This fixes various strange behaviors such as tooltips not disappearing for multiple seconds after the cursor leaves, and modals not having open animations.
- Fixes the broken header on mobile (bad layout, logout button displaying without appstore enabled)
- Fixes a bug where if a user searches a query, then presses the "search" entry in the header menu, it would break the layout.
- Fixes a couple small bugs concerning expanded layout behavior that made it past the initial PR.
- A few smaller obscure bugs.


mbwatson:
@frostyfan109 awesome! is this what's currenty deployed in your namespace?

frostyfan109:
> @frostyfan109 awesome! is this what's currently deployed in your namespace?

I'm deploying it now

mbwatson (changes requested):
what a huge upgrade to the helx ui. excellent job piecing all this together. i think it's awesome.

one thing: i'm seeing lots of this error: `Uncaught (in promise) TypeError: this._flags is undefined` from mixpanel.

mbwatson (approved):
:+1:

PR #203, opened 2022-07-18
CDEs & Tab Search

frostyfan109:
## Overview
Adds CDE UI components as specified in this issue: https://github.com/helxplatform/dug/issues/221.
- Adds CDEs tab in modal (detail box). Lists CDEs' names, descriptions, related concepts.
- Related concepts can be clicked to navigate to a new detail box or to navigate to a new dug search.
- Displays CDEs under studies tab in card (summary box) with modified UI.
- Also removes knowledge graph tab from card (summary box) in an effort to make the card more of a "summary" with reduced technical details.
- Adds search functionality to studies, CDEs, and knowledge graph tabs within the detail box.
- Adds abort controllers to API requests

### Search
The studies, CDEs, and knowledge graph tabs inside the concept modal all now come with search boxes at the top for easier navigation. Searches also highlight the relevant terms that resulted in the hit, which is something that is being considered as well for top-level dug searches.
- CDE searches will reveal highlighted related concepts if hidden by the `Show more` button.
- Studies searches will reveal variables that contain highlighted terms if the study is collapsed.
- The knowledge graphs tab doesn't obscure any information so no special behavior is necessary.

Search is implemented using Lunr.js, which is a lightweight in-browser full-text search engine inspired by Solr/Lucene.
- Search is built to be easily reusable within the app. This adds two hooks: a thin Lunr wrapper and a further wrapper `useLunrSearch` which facilitates use within a component.
- This also adds `DebouncedInput` for debounced searches. This could possibly be used to replace the dug search box in the future as well.

### Abort controllers
Abort controllers are handy for cancelling stale API requests should the UI no longer require the result. For example, if I open the details box for a concept, it's going to fire a few requests to load associated studies, CDEs, and knowledge graphs. Perhaps the requests are going particularly slowly and I'm getting impatient, so I close the details box before all the requests finish. The UI no longer cares about these requests, and their completion will effectively be ignored. So they should be cancelled.
- This was somewhat irrelevant before but should still be in place since it's good practice.
- Since loading related concepts fires so many tranql requests, it's particularly important to make sure they get cancelled if the user closes the concept modal.

### Search context changes
The search context now supports the method `searchSelectedResult`, which takes a concept name and id and tries to find the associated dug result (done through performing a search against `/search` and sifting through results). This lets you set `selectedResult` to a concept whose data is not loaded in the app state. This is used in CDE related concepts since we only have the tranql result for a concept, not the dug result, which is required by `setSelectedResult`.

`selectedResult` is now decorated with a few extra fields.
- `previousResult` is automatically added to indicate the previous `selectedResult` value. This lets the UI add a back arrow to navigate to previous results when, for example, you open a related concept in a CDE. 
- `loading`  is used for asynchronous selection of a concept (e.g. used in searchSelectedResult while searching)
- `failed` is used to indicate that the attempted loading of a concept was not found or for some other reason could not be loaded (e.g. a tranql concept isn't indexed in dug in searchSelectedResult)

### Fixes
Fixes https://github.com/helxplatform/dug/issues/238 where some concept card/modal titles were `xxx (disease) (disease)`

### Current related concepts limitations
The tranql logic responsible for the CDE "Related concepts" will need to be updated in the future to leverage general tranql queries using `named_thing` (once this is added, of course).
- Right now, it performs an enormous number of queries against the most frequent concept types (currently, it makes one query per CDE per type `disease`, `phenotypic_feature`, `anatomical_entity`, `biological_process`).
- So for concepts with, say 10 CDEs, 40 tranql requests will be fired simultaneously, and we'll also only get results for these 4 specific concept types, neither of which is ideal.

As well, there is some funny stuff happening with the internal tranql schema that doesn't allow to query against `clinical_modifier`, despite being in the schema. Since concepts linked to a clinical modifier are "associated" with studies, and thus will be indexed in dug, there's no way currently to tell ahead of time if a related concept is indexed in dug or not without performing a dug search for every single related concept, which obviously isn't reasonable.
- This isn't a big deal, and there is currently a UI in place for redirecting users to other concepts should they navigate to a related concept that isn't indexed in dug.
- It would be nice however if perhaps the "Open" feature on a related concept was instead disabled on non-indexed concepts so that users don't waste their time.


src/components/search/concept-modal/concept-modal.js
```diff
@@ -60,17 +89,159 @@ export const ConceptModalBody = ({ result }) => {
...
+      
+      getVars()
+      getCdes()
+      getKgs()
+    }
+  }, [fetchKnowledgeGraphs, fetchStudyVariables, fetchCDEs, result, query])
 
+  useEffect(() => {
+    return () => {
+      // On unmount, cancel all pending fetch requests.
```
mbwatson, line 206: :+1:

src/components/search/concept-modal/concept-modal.js
```diff
@@ -60,17 +89,159 @@ export const ConceptModalBody = ({ result }) => {
...
+        }
+        const relatedConcepts = {}
+        if (cdeData) {
+          const cdeIds = cdeData.elements.map((cde) => cde.id)
+          await Promise.all(cdeIds.map(async (cdeId, i) => {
+            relatedConcepts[cdeId] = await loadRelatedConcepts(cdeId)
+          }))
+        }
+        setCdes(cdeData)
+        /** Note that relatedConcepts are *TranQL* concepts/nodes, not DUG concepts. Both have the top level fields `id` and `name`. */
```
mbwatson, line 171: :+1:

src/components/search/concept-modal/tabs/cdes/cdes.js
```diff
@@ -0,0 +1,63 @@
...
+  /** CDEs have loaded, but there aren't any. */
+  const failed = useMemo(() => !loading && !cdes, [loading, cdes])
+
+  const docs = useMemo(() => {
+    if (!loading && !failed) {
+      return cdes.elements.map((cde) => ({
+        id: cde.id,
+        name: cde.name,
+        description: cde.description,
+        // Lunr supports array fields, though it expects the user to tokenize the elements themselves.
```
mbwatson, line 27: :+1: :+1:

src/components/search/concept-modal/tabs/cdes/related-concepts/related-concept-tag.css
```diff
@@ -0,0 +1,28 @@
...
+.related-concept-popover-overlay .ant-popover-title {
+    padding: 8px 12px;
+    /* text-align: center; */
+    text-transform: capitalize;
+}
+.related-concept-options {
+    padding-bottom: 4px !important;
+}
+.related-concept-options > .ant-menu-item {
+    margin: 0 !important;
```
mbwatson, line 17: are we missing something with antd? all these `!important`s don't feel right

src/components/search/context.js
```diff
@@ -23,6 +19,10 @@ const tempSearchFacets = [
   'LOINC',
   'PhenX',
 ].sort((f, g) => f.toLowerCase() < g.toLowerCase() ? -1 : 1)
+// The search_var endpoint returns studies (DbGaP, etc) and CDEs.
```
mbwatson, line 22: :+1:

src/contexts/environment-context.js
```diff
@@ -17,7 +17,10 @@ axios.defaults.xsrfCookieName = 'csrftoken';
 axios.interceptors.response.use(function (response) {
   return response;
 }, function (error) {
-  if (error.response.status === 403) {
+  if (error instanceof CanceledError) {
```
mbwatson, line 20: :+1:

src/hooks/use-lunr-search.js
```diff
@@ -0,0 +1,64 @@
...
+    docs
+}) => {
+    const initIndex = useCallback(function () {
+        this.ref(ref)
+        fields.forEach((field) => {
+            this.field(field)
+        })
+        this.metadataWhitelist = ["position"]
+        // Lunr has really bad docs for how the engine tokenizes text.
+        // As far as I can tell, Lunr strictly uses pattern-based tokenization.
```
mbwatson, line 26: :+1:

src/hooks/use-lunr-search.js
```diff
@@ -0,0 +1,64 @@
+/**
```
mbwatson, line 1: excellent idea

mbwatson (approved):
another awesome upgrade to the UI that i think will be well received. good job with the code, as well as all the the useful comments. the search feature is very nice, and i learned a few things while working through it! good job!

YaphetKG (approved):
Neat stuff !

src/components/search/concept-modal/tabs/cdes/related-concepts/related-concept-tag.css
frostyfan109: Good catch. This part is a little hacky. Basically, what we're going for here is the appearance of a dropdown menu without using a dropdown (this uses a `Popover` instead, very similar components though).

antd's `Dropdown` component expects the `overlay` prop to be a `Menu`. It then takes the provided menu and styles it accordingly to give the distinct dropdown menu look. On the other hand, the `Popover` component can take anything as its content and has somewhat different styling.
- You lose the dropdown menu look when using a Popover, so you have to manually override the styling of `Menu` with `important!` rules (although you could probably drop these if more specific selectors were specified here). 

The reason this is using a Popover, even though it is literally mimicking a Dropdown is because:
1. Dropdown doesn't support `right` alignment. Technically, you can still use it (since it has a shared API with Popover) but you lose the overlay arrow since it isn't intended to be used.
2. A dropdown doesn't also have a title. It is *just* a menu. Not really a deal breaker but an added perk of using a popover.

PR #204, opened 2022-07-19
View/Page Titles

frostyfan109:
Adds titles to each UI view instead of just the static `HeLx UI`. The format used is `{name} · HeLx UI`, with extra bullets being inserted when relevant, e.g. searches use `Search · {query} · HeLx UI`.


mbwatson (approved):
:+1: looks good

PR #208, opened 2022-09-27
Fix timout and depth

Hoid:


Hoid:
Somehow I never fixed the timeout and depth from testing before pushing to develop, so I'm fixing that 😅

PR #212, opened 2022-10-06
Remove package.json version

Hoid:


Hoid:
Forgot to remove this in the last PR

waTeim:
So the idea here is that since appstore is not to every be published to npmjs, and it might come into conflict with the versions in other places, why maintain it?

frostyfan109 (approved):

PR #224, opened 2022-10-20
Variable View

frostyfan109:
This adds the variable view layout which distinguishes concept vs variable search results. See: https://github.com/helxplatform/dug/issues/230 for additional info.


mbwatson (approved):
:tada: i know this was a beast of a merge, but it all seems to work as intended. i have a few suggestions that you can think about and/or implement as you see fit. i think the Start Over and forward/backward arrow buttons could be clarified. maybe something like "reset zoom" and "undo"/"redo" for the forward and back buttons could work well.

on an unrelated note, there are going to be issues with making this new view accessible. if HeLx ever gets deployed on a dot-gov domain, abiding by WCAG standards are necessary. i don't think this is something to slow down development now, but it _is_ something that should be on the radar.

frostyfan109:
> Start Over and forward/backward arrow buttons could be clarified. maybe something like "reset zoom" and "undo"/"redo" for the forward and back buttons could work well.
I will experiment with adding textual hints before merging.

> abiding by WCAG standards is necessary.
Yeah,  it would be good to address this earlier than later. I'm not qualified to ensure we have WCAG compliance though.

mbwatson (approved):
look good!

PR #228, opened 2023-01-23
fix variable-view package-lock

frostyfan109:
basically just regenerated the package-lock which ended up broken when variable-view was merged


Hoid (approved):
LGTM, but let's wait for it to build successfully before merging

PR #230, opened 2023-01-31
fix/Update Jenkins file to use CCV

joshua-seals:


Jenkinsfile
```diff
@@ -52,27 +68,29 @@ spec:
         REPO_NAME="helx-ui"
         COMMIT_HASH="${sh(script:"git rev-parse --short HEAD", returnStdout: true).trim()}"
         IMAGE_NAME="${REGISTRY}/${REG_OWNER}/${REPO_NAME}"
-        TAG1="$BRANCH_NAME"
```
Hoid, line 55: you'll need these, since they're referenced down below

Jenkinsfile
```diff
@@ -1,10 +1,12 @@
 library 'pipeline-utils@master'
 
+CCV = ""
+CURR_TIMESTAMP = ""
```
Hoid, line 4: you can remove this CURR_TIMESTAMP, I think I was using it a while back but now it's just a local variable below, so the global is unnecessary

Jenkinsfile
Hoid: oops nevermind, misread this, this is fine

Hoid (approved):

PR #232, opened 2023-02-07
Helps when running workspaces locally

frostyfan109:
Automatically points the UI to localhost:8000 to look for a running appstore instance


frostyfan109:
Note: even with this fix, you still need to disable the cross-origin security policy on your browser in order to properly interface with appstore.

joshua-seals (approved):

PR #239, opened 2023-02-17
Fix commit-msg git hook again (real this time)

Hoid:


Hoid:
This should have the changes to the commit-msg hook in it AND a correct commit message to trigger the right CCV value on merge.

waTeim (approved):

PR #241, opened 2023-02-17
External app registry

pj-linebaugh:
Add support to use an app-registry.yaml file from a git repo outside of the Tycho code.


bin/populate_env
```diff
@@ -51,6 +51,6 @@ echo "$template" | sed \
   -e "s/%TRANQL_ENABLED%/$tranql_enabled/" \
   -e "s/%TRANQL_URL%/$tranql_url/" \
   -e "s/%DEPLOYMENT_NAMESPACE%/$deployment_namespace/" \
-  -e "s/%DOCKSTORE_BRANCH%/$dockstore_branch/" \
+  -e "s+%DOCKSTORE_APP_SPECS_DIR_URL%+$dockstore_app_specs_dir_url+" \
```
Hoid, line 54: what's with the `+` at the beginning and end? is that supposed to be there?

Hoid: also, the rest of these all have a slash at the end after the var substitution. do we need that on this line too?

pj-linebaugh: The '+' is used as a replacement for '/' since the env variable contains forward slashes and is 'expanded'.  See [this page](https://www.cyberciti.biz/faq/how-to-use-sed-to-find-and-replace-text-in-files-in-linux-unix-shell/) as a reference for it being used.

waTeim (approved):

PR #243, opened 2023-02-27
Fixes app documentation link

frostyfan109:
- Fixes the broken documentation link in app "about" panel
- Adds a line break between app description and documentation link


mbwatson (approved):
:+1:

PR #244, opened 2023-03-01
Add useTitle hook to views

frostyfan109:
This was originally supposed to add a higher-order component used by all view components, but with the only configuration needed on the view-level being the title and the need for dynamic titles in certain views, it didn't make sense as of right now. If we need more configuration for views in the future, we may want to revisit this idea.

For now, this PR just adds a useTitle hook that each view should use to set the document title when the view is active.

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -1,3 +1,4 @@
+export * from './view'
 export * from './support'
 export * from './not-found'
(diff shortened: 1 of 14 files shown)


mbwatson (approved):
:+1: clever implementation

PR #245, opened 2023-03-03
fix: critical bug related to mishandling stale data

frostyfan109:
Fixes bug that would occur if a user navigated to a related CDE concept and then quickly returned to the original concept. Resulted in a page crash if the concept modal was not closed before the requests resolved.


mbwatson (approved):
:+1:

PR #247, opened 2023-03-22
Add toggleable ui card/modal tabs. Separate studies/cdes tab in result card into two standalone tabs

frostyfan109:
- Replace `REACT_APP_TRANQL_ENABLED` with a more general `REACT_APP_HIDDEN_RESULT_TABS` which is a comma-delimited list to toggle the visibility of tabs in the card/modal for search results
- Separate the card tab for studies/cdes into two standalone tabs, one specifically for studies and one for cdes. Keeps the exact same UI for both, just separated.


gaurav:
Thanks, @frostyfan109! I tried running this without any configuration changes and got the following error:

```
Unhandled Rejection (TypeError): context.hidden_result_tabs is undefined

src/contexts/environment-context.js:91
context.hidden_result_tabs = context.hidden_result_tabs.split(',').map((tab) => tab.trim())
```

I think you need to set a default for this property so it works out-of-the-box. Could you please update the code to do that? If something is wrong at my end, please let me know!

frostyfan109:
@gaurav
It should work out of the box assuming you regenerate your `env.json` file using the `populate_env` script or manually add in the `"hidden_result_tabs": ""` key to your json file. The idea is that defaults are stored in the `.env.sample`, you copy these values into your own `.env` file, configure them however you like, and then generate the `env.json` file. The original practice was that we didn't add default values baked directly into the code to catch missing `env.json` keys unless the environment variable relates is for a development feature (i.e. it isn't used in a production deployment). That's rather inconsistent now though.

It may be worth creating a ticket to address adding defaults to everything in case they're missing from your `env.json`.

.env.sample
```diff
@@ -8,14 +8,14 @@ REACT_APP_HELX_SEARCH_URL='https:\/\/heal-dev.apps.renci.org'
...
 REACT_APP_ANALYTICS=''
-# Enables the TranQL tab in search concept popups
-REACT_APP_TRANQL_ENABLED='true'
 # Points the UI to TranQL. TranQL often runs on /tranql
 REACT_APP_TRANQL_URL='https:\/\/heal-dev.apps.renci.org\/tranql\/'
 # Brand setting. <braini, catalyst, heal, scidas, eduhelx, helx>
 REACT_APP_UI_BRAND_NAME='catalyst'
 # Hide support page sections if necessary. <'community', 'documentation'>. Comma-delimited string, e.g. "community,documentation"
 REACT_APP_HIDDEN_SUPPORT_SECTIONS=''
+# Hide tabs in the result modal/cards. <'overview', 'studies', 'cdes', 'kgs', 'tranql', 'robokop'>. Comma-delimited string, e.g. "tranql,robokop"
```
gaurav, line 17: I would add a note here to clarify that this only removes these tabs from the modal, not from the cards on the main concept search page (as far as I can tell -- please correct me if I'm wrong!).

gaurav (approved):
Thanks for that information, @frostyfan109 -- re-running populate_env.sh worked for me! I was able to test this with `cdes,robokop` which is probably what we'll need for BDC Dug. I've left a comment suggesting that you clarify that this will turn off CDEs and Robokop in the modal, but not in the main concept search page (i.e. even with CDEs turned off in the modal, I still see it on the main concept search page). If this is supposed to turn off CDEs everywhere, please let me know if there's something I'm doing wrong with the configuration. Thanks so much for your work on this!

frostyfan109:
@gaurav Yeah, I think my intention was to have the option hide the tab in both concept modal/cards, but I must have forgotten to make it hide the tabs in cards, will fix it.

PR #249, opened 2023-04-18
fix: Fix broken env script

frostyfan109:
Change @pj-linebaugh made I think uses wrong sed substitution, populate_env now crashes and can't generate an env file. I've gone ahead and just removed the option since it isn't useful except in the different populate_env present in appstore, and will be gone soon anyways with decoupling changes.


frostyfan109:
Thought the new dockstore option was causing problems with the env script, but it turns out it was something unrelated. I'll close this PR since the same changes are slated to be included in the decoupling changes. merging this wouldn't hurt anything, but it's unnecessary.

PR #256, opened 2023-05-18
Redo workspaces api/context coupling and responsibilities

frostyfan109:
When the workspaces API was initially implemented, user-state and session timeout logic was baked directly into the API implementation. This worked, and there was some decent reasoning for why it was convenient. But ultimately, it was just bad practice and introduced way too much needless complexity into what should be a somewhat pure wrapper class. So I've gone ahead and moved both of these into the context itself, where they probably should have been to begin with.

Also adds an event emitter into the API for a more streamlined event system in anticipation of websocket support down the line. Changes user state event to `userStateChanged(userState: User | null)`. Adds new event, `apiRequest(methodName: string, promise: Promise<any>)`.
- apiRequest is emitted immediately after the API method is invoked.  
- Both apiRequest and apiError may be emitted for a single API request invocation.
- This is really only intended for internal use in the context for tracking session timeout.

Also changes mentions of whitelist to approval.


pj-linebaugh (approved):
Looks good AFAIK.  Lots of changes there.  Let's get a new image created and test it out

frostyfan109:
> Looks good AFAIK. Lots of changes there. Let's get a new image created and test it out

@pj-linebaugh Working on bundling websocket changes into this PR, seems fitting to include them. I can branch off this and do it in two prs if we want to release this now though. It will still take a few more days to finalize the websockets.

pj-linebaugh:
Looks like you already added in the websockets code, which is fine with me.

PR #263, opened 2023-07-21
Feature/dug 0.9 UI

frostyfan109:
- Collapsible histogram
- Count indicators for studies/cdes
- Dynamic title/description in meta tags. This should make it so that when you paste a public link (like [heal.renci.org](http://heal.renci.org/) or [search.biodatacatalyst.org](http://search.biodatacatalyst.org/)) into a smart app (e.g. slack), it should display actually relevant information about the website, instead of saying some gibberish about create-react-app.
  - We’ll need to have people deploying configure this info though.
- Improve handling if a search returns no results, also fixes the error message if a search api call fails to go through


gaurav (approved):
This looks awesome, @frostyfan109 -- I particularly like the counts, spinners and concept links! I had to make a minor change to make this work on my computer (c92a2b7a7a3a3abd7c35c05cebd44847ce8f0897) but otherwise this looks great to me. Thank you!

PR #264, opened 2023-07-24
Remove packages with CVEs

ptlharit2:
Updating the base image to this slimmer version give the basic Node libraries and most of the same linux ones. Main benefit is its a smaller size and fewer vulnerabilities


frostyfan109 (approved):
👍

PR #267, opened 2023-08-18
Update concept-card.js

YaphetKG:
Removing study counts from tab title based on convo here:  https://renci.slack.com/archives/C012E0H6LP5/p1692294112429759


frostyfan109:
This is m removing the counts in the cards, not the modal. I think the slack conversation pertains to the modal counts instead.

PR #269, opened 2023-09-01
Update support page to include HEAL-specific documentation

gaurav:
This PR updates the Dug Support page to have a `heal-support` section, which can be hidden by adding `heal-support` to the configuration variable `hidden_support_sections`. Note that if this is merged, the Helm charts for all the Dugs will need to be updated, so that HEAL Dug's `hidden_support_sections` is set to `community,documentation` and all the other Dugs are set to `heal-support` to prevent showing all three sections.

This is what the support page should look like for HEAL:
Screenshot 2023-09-01 at 3 48 39 PM


mbwatson (approved):
nice!

frostyfan109 (approved):
👍

PR #270, opened 2023-09-05
Use context.meta.title in footer and mobile menu header

mbwatson:
this PR replaces uses the context value `context.meta.title` in place of "HeLx" in two places: the footer and the mobile menu.

note this approach mimics what's already in place in `view.tsx` for the window title.


gaurav (approved):
This looks great to me! Should we add `target="_blank"` to the new `a`s so that they open in a new window?

PR #272, opened 2023-09-15
Remove counts from concept modal and minor lint fixes

gaurav:
This PR removes the counts from the tabs in the concept modal, as brought up in https://renci.atlassian.net/browse/DUG-348. As far as I can tell, the code being deleted here is identical to the code in src/components/search/concept-card/concept-card.js, so I have no idea why the counts are off, but this should hide the oddness until we can properly fix and close https://renci.atlassian.net/browse/DUG-348.

I ran into some linting issues and had to fix them to try this out:
1. `npm run lint` required adding `rel=noreferrer` for the two `target='_blank'` links added in PR #270, with a link to
https://mathiasbynens.github.io/rel-noopener/#recommendations by way of an explanation. I've done that, but I wonder if `target='_blank'` is a bad idea for some reason?
2. Removed one unnecessary semicolon.

Supercedes PR #267

### DIFF
--- a/src/components/layout/layout.js
+++ b/src/components/layout/layout.js
@@ -95,7 +95,7 @@ export const Layout = ({ children }) => {
       {
         context?.brand === 'heal'
         ? <Footer style={{ textAlign: 'center', paddingTop: 0 }}>
-          HEAL Semantic Search is powered by Dug, an open source semantic search developed
+          NIH HEAL Semantic Search is powered by Dug, an open source semantic search developed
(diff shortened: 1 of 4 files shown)


mbwatson (approved):
yep, good catch! we forgot noreferrer, so this now looks good.

regarding the different study counts in the card vs the modal, i've noticed a few occasional instances of different study responses across several request attempts before. it was on the bdc instance, but maybe this is a case of the same phenomenon. i haven't been able to reliably reproduce this, though. i'll try to get a screenshot next time i notice it happening.

frostyfan109:
@gaurav `target="_blank"` isn't problematic, it's just a security issue in some of the older browsers if you omit the rel part when using it.

frostyfan109 (approved):
👍

YaphetKG (approved):

PR #275, opened 2023-09-19
Add support for custom branded favicons

frostyfan109:


YaphetKG (approved):

gaurav (approved):
I can't test this locally, but it is working correctly at https://groupe.apps.renci.org/, so it should be correct!

frostyfan109:
I've tested locally, to make sure everything is working still. It's just that branding favicons won't work locally. This is because for development, we use the webpack dev server, so we do:
- index.ejs -> index.html (no intermediate step)
For production, webpack does: 
- index.ejs -> index_template.html
Then the `start_server` script generates an env.json file, and also generates an `index.html` using `index_template.html` and your environment.

PR #278, opened 2023-10-10
Alter readiness check to use the is_ready API endpoint 

waTeim:
Add support for checking the readiness of an App in a different way.  Previously, the UI (splash screen) would try to 
connect the main interface of the App.  With this change, it instead checks against a new API endpoint. `/v1/instances/{sid}/is_ready/`.  The result will be false until the pod underlying the App passes its readiness probe.

### DIFF
--- a/src/contexts/instance-context.js
+++ b/src/contexts/instance-context.js
@@ -30,7 +30,7 @@ export const InstanceProvider = ({ children }) => {
         // let ready = false;
         const decoded_url = decodeURIComponent(app_url);
         const executePoll = async () => {
-            const isAppReady = await api.getAppReady(app_url);
+            const isAppReady = await api.getAppReady(decoded_url);
             if (isAppReady) {
                 let newActivity = {
                     'sid': sid,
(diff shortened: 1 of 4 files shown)


pj-linebaugh (approved):
looks good to me.

frostyfan109 (commented):
looks good in general. couple notes:
- fetchOptions should get passed as the second arg of the axios.get, right now it's not being used.
- Also see that there's already a `getAppReady` method in the API. You should just replace that one with the new one, and make sure it's still working wherever the existing method was being used. Also make sure to update the `getAppReady` declaration in the IWorkspacesAPI interface, or you'll get errors.

waTeim:
updated to conform to suggestions

frostyfan109:
> updated to conform to suggestions

Looks good, only gripe is the api call should be try caught by the user, not inside the call itself. Putting it inside the api call can mess up some of the error handling

waTeim:
Hmm yeah true, but I'll also say that in both of the calling contexts where it is invoked, the code treats an error the same as false and simply blindly retries the call after a pause.

frostyfan109:
Yeah but it’s about code consistency. And also like I said, the error handling in the code for any APIRequest method breaks if you catch errors inside the request, so it literally breaks the code in edge cases to do that

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

waTeim:
modded the try-catches too.

frostyfan109 (approved):
very cool !

PR #279, opened 2023-10-16
BUG: adjust workspace tablist to avoid jyttery effect on Windows

hina-shah:
Resolve HELX-369
Adjust the workspace tab group so that it does not cause confusion on Windows web browsers. Still need to test it on Windows though.


waTeim:
I note that these days you don't need to use buildx (historical) and simply the following s good enough

    docker  build --platform=linux/amd64

waTeim (approved):

PR #283, opened 2023-10-26
Creating Release for UI

hina-shah:
breaking:

This adds:
- Readiness probe on Kubernetes Cluster
- Non-root image for UI
- A few bugs related to UI experience
- Dug changes


joshua-seals (approved):
yolo - this should never be so big again. It completely invalidates the code review process

frostyfan109 (approved):

Dockerfile
```diff
@@ -1,16 +1,16 @@
 ###################
 # Build environment
 ###################
-FROM node:18.16-alpine AS builder
+FROM node:lts-bullseye-slim AS builder
```
pj-linebaugh, line 4: I think using versions is better so that we know for sure what image tag was used.

pj-linebaugh (approved):

PR #293, opened 2023-12-12
Updating login page/signup page text

hina-shah:
Updating the logiin/signup page text to be more descriptive.


public/index.ejs
```diff
@@ -36,7 +36,7 @@
     <title><%= isDevelopment ? "HeLx UI" : "%META_TITLE%" %></title>
   </head>
   <body>
-    <noscript>You need to enable JavaScript to run this app.</noscript>
```
frostyfan109, line 39: I don't understand why you changed this? The website is built on React (which is Javascript), and the content of noscript displays when Javascript is disabled. So it's more or less just an informational error message for the user that the website is unusable unless they enable JavaScript.

Feel free to improve the error message but changing this to HeLx UI would be confusing.

src/views/workspaces/login/login.js
```diff
@@ -183,13 +183,14 @@ export const WorkspaceLoginView = withAPIReady(({
...
-                            has broad reaching benefits and can be applied in many domains.
+                            <a href="https://helxplatform.github.io/">HeLx</a> empowers researchers in domains from plant genomics to neuroscience to work with their preferred tools and apps in the cloud at scale.
+                            Its ability to empower researchers to leverage advanced analytical tools without installation or other infrastructure concerns has broad reaching benefits and can be applied in many domains.
+                            <br />&nbsp;<br />
+                            The HeLx Workspaces provide a wide array of data science tools for use by these researchers. Through the Workspaces, users explore and interact with analytic tools and data to support scientific discovery.
+                            <br />&nbsp;<br />
+                            Please sign in with one of the following options to access the HeLx Workspaces.
                         </Paragraph> : null
                     }
-                    logo={ <AppstoreOutlined /> }
```
frostyfan109, line 192: Why'd we delete this?

src/views/workspaces/login/signup.js
```diff
@@ -182,7 +184,6 @@ export const WorkspaceSignupView = withSocialSignupAllowed(({
                             </Paragraph>
                         </div>
                     }
-                    logo={ <AppstoreOutlined /> }
```
frostyfan109, line 185: Again confused why we're deleting this logo; are we not happy with how it looks or is it visual clutter or something else? I'd prefer we replace it if we just don't like the appearance.

frostyfan109 (changes requested):

public/index.ejs
hina-shah: This text was showing up on Google Search. We can change it.

src/views/workspaces/login/signup.js
hina-shah: The hope was to make that logo to be something else and not be appstore specific.

frostyfan109 (approved):

PR #296, opened 2024-01-25
Add new event parameters for purposes of tracking MRR

frostyfan109:
Adds resultRank parameter to interaction with search result. And adds tracking for card tab clicks.


hina-shah:
Hi @frostyfan109 : Do you mind changing the PR with some description please? It would be helpful to understand what the PR is doing in future. Thanks!

PR #302, opened 2024-04-11
Preparing action merge

pchachicho:
Added the new paths for helx-github-actions to helx-ui. Some of the code is commented out for testing purposes.


joshua-seals:
Patrick, you have to un-comment those restrictions in the dev and master workflows. We had a release that was not supposed to occur. Definitely don't want to do that when we merge.

PR #308, opened 2024-06-20
ENH: Adding study linkages to CDE tab in concept card

hina-shah:
Showing studies that use CDEs displayed in the concept card of concept search.

### DIFF
--- a/src/components/search/concept-modal/concept-modal.js
+++ b/src/components/search/concept-modal/concept-modal.js
@@ -214,7 +214,7 @@ export const ConceptModalBody = ({ result }) => {
             return `\
       SELECT publication->study
       FROM "/schema"
-      WHERE cde="${cdeId}"`
+      WHERE publication="${cdeId}"`
           }
           const tranqlUrl = context.tranql_url
           const controller = new AbortController()
--- a/src/components/search/concept-modal/concept-modal.js
+++ b/src/components/search/concept-modal/concept-modal.js
@@ -78,6 +78,7 @@ export const ConceptModalBody = ({ result }) => {
   const [studies, setStudies] = useState([])
   const [cdes, setCdes] = useState(null)
   const [cdeRelatedConcepts, setCdeRelatedConcepts] = useState(null)
+  const [cdeRelatedStudies, setCdeRelatedStudies] = useState(null)
   const [cdesLoading, setCdesLoading] = useState(true)
 
   /** Abort controllers */
@@ -100,7 +101,7 @@ export const ConceptModalBody = ({ result }) => {
   const tabs = {
     'overview':     { title: 'Overview',         icon: <OverviewIcon />,         content: <OverviewTab result={ result } />, },
     'studies':      { title: studyTitle,         icon: <StudiesIcon />,          content: <StudiesTab studies={ studies } />, },
-    'cdes':         { title: cdeTitle,           icon: <CdesIcon />,             content: <CdesTab cdes={ cdes } cdeRelatedConcepts={ cdeRelatedConcepts } loading={ cdesLoading } /> },
+    'cdes':         { title: cdeTitle,           icon: <CdesIcon />,             content: <CdesTab cdes={ cdes } cdeRelatedConcepts={ cdeRelatedConcepts } cdeRelatedStudies={cdeRelatedStudies} loading={ cdesLoading } /> },
     'kgs':          { title: 'Knowledge Graphs', icon: <KnowledgeGraphsIcon />,  content: <KnowledgeGraphsTab graphs={ graphs } />, },
     'explanation':  { title: 'Explanation',      icon: <ExplanationIcon />,      content: <ExplanationTab result={ result } /> },
     'tranql':       { title: 'TranQL',           icon: <TranQLIcon />,           content: <TranQLTab result={ result } graphs = { graphs } /> }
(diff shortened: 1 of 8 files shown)


gaurav (approved):
The code looks great! I have some suggestions for the UX (and one very minor code suggestion), but I'm not sure if we want to do that in this PR or if it would be better to do that in a separate PR. My suggestions would be:
1. I don't think related studies is as important as the description; instead, I'd reduce the font size a bit so it was similar in size to the "Related concepts".
2. The related studies list can be pretty long -- I would either display only the first three entries followed by a "show more" button or collapse this entire section.

Also I notice that the Highlighter can't be used to search by Study Title -- that would be useful to have, but not critical for this PR. I think we should open a ticket for that and work on it separately.

src/components/search/concept-modal/concept-modal.js
```diff
@@ -207,7 +208,41 @@ export const ConceptModalBody = ({ result }) => {
             }
           )
         }
+
+        const loadRelatedStudies = async (cdeId) => {
+          const formatCdeQuery = () => {
+            return `\
```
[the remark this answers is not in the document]
hina-shah: I agree it's a short query, but I think it helps with readability.

hina-shah:
@gaurav : I've pushed requested changes to the related study list.

gaurav:
@hina-shah Thanks for working on this, Hina! I found a bug in the TranQL query for CDEs and fixed that, and then made a few minor improvements to the studies display. Have a look and let me know what you think -- and feel free to remove any of my "improvements" if you disagree! My previous approval still stands :)

hina-shah:
Thanks @gaurav !

src/components/search/concept-modal/tabs/cdes/cde-item.js
```diff
@@ -19,7 +23,57 @@ const Section = ({ title, children }) => (
...
+    //   analyticsEvents.studyLinkClicked(study.c_id)
+    // }
+    const [showMore, setShowMore] = useState(false)
+
+    if (!relatedStudySource) return (
+        <Text style={{ fontSize: 13, color: "rgba(0, 0, 0, 0.45)" }}>We couldn&apos;t load any related studies.</Text>
+    )
+
+    // Sort the related studies.
+    const relatedStudies = relatedStudySource.sort((a, b) => { return b.c_id < a.c_id});
```
frostyfan109, line 38: This is a bit nitpicky, but typically "source" variables are already sorted in the desired order. Ideally, we'd be sorting them in the fetcher method, CdeItem would also be alright.

src/components/search/concept-modal/tabs/cdes/cde-item.js
```diff
@@ -19,7 +23,57 @@ const Section = ({ title, children }) => (
     </Space>
 )
 
-export const CdeItem = ({ cde, cdeRelatedConcepts, highlight }) => {
+const RelatedStudiesList = ({relatedStudySource}) => {
```
frostyfan109, line 26: Could you move this into a separate folder, like how `./related-concepts` is?

src/components/search/concept-modal/tabs/cdes/cde-item.js
```diff
@@ -19,7 +23,57 @@ const Section = ({ title, children }) => (
...
+            <List
+                size="small"
+                dataSource={showMore ? relatedStudies : relatedStudies.slice(0, SHOW_MORE_CUTOFF)}
+                renderItem={(study) => (
+                    <List.Item key={study.c_id}>
+                    <List.Item.Meta
+                        size="small"
+                        description={
+                            <>
+                            <font size="-1"><a href={study.c_link} target="_blank" rel="noopener noreferrer">{study.c_id}</a></font>: {study.c_name}
```
frostyfan109, line 55: don't use `<font>`, typically you'd just set the font-size as pixel units in css/inline style. If you really want to calculate the font size based on its inherited value, you can use `font-size: calc(100% - 1px);` I think, but it's completely fine to just hardcode it to 1 px less than whatever the inherited value is.

src/components/search/concept-modal/tabs/cdes/cde-item.js
```diff
@@ -19,7 +23,57 @@ const Section = ({ title, children }) => (
...
+                            </>
+                        }
+                    />
+                    </List.Item>
+                )}
+            />
+            { relatedStudies.length >= SHOW_MORE_CUTOFF && (
```
frostyfan109, line 62: Think this should be >, not >=. Also, the content is triple indented for some reason.

src/components/search/concept-modal/tabs/cdes/cde-item.js
```diff
@@ -3,12 +3,16 @@ import { List, Collapse, Typography, Space, Button } from 'antd'
 import { ExportOutlined } from '@ant-design/icons'
 import _Highlighter from 'react-highlight-words'
 import { RelatedConceptsList } from './related-concepts'
+import Highlighter from 'react-highlight-words'
```
frostyfan109, line 6: This is imported here twice using different default names

frostyfan109 (changes requested):

src/components/search/concept-modal/tabs/cdes/cde-item.js
hina-shah: I'm going to leave this as it is for now.

frostyfan109 (approved):

gaurav:
@hina-shah If @frostyfan109 approves of this PR, I think it's good to merge -- I'd love to see what it looks like in Dev and to get more feedback from other HEAL folks!

PR #309, opened 2024-06-29
changed console.log to openNotification.. when refering to API calls

pchachicho:


frostyfan109:
@pchachicho sorry it took so long for me to get to this. this pr is on the right track, but there are a lot of other usages of the workspaces API throughout the UI components ([for reference](https://github.com/search?q=repo%3Ahelxplatform%2Fhelx-ui+%22await+api%22&type=code)). Some of these don't need error messages (e.g. in app-card, we don't need to show a message if the app fails, since that's handled elsewhere) but the majority most do

PR #311, opened 2024-07-22
Variable view rewrite & consistent data source coloring

frostyfan109:
- Rewrite/cleanup variable view w/context so that we can support filters and other features that weren't otherwise feasible
- Add infinite scroll to variable view
- Fix study data source (e.g., HEAL studies vs Non-HEAL studies) coloring
- Group by study/collapse into variable list
- Add filters:
  - Subsearch filter
  - Data source filtering
  - Sort by score/data source, with ascending/descending options


joshua-seals:
RUN npm clean-install is causing the builds to fail, just fyi

gaurav:
I can't review the many code changes in this PR, but the end results look great! Some thoughts:
1. I think the subsearch filter is a bit too fuzzy: I tried filtering with "heart" and ended up matching a ton of "healing"s and one "part". But I LOVE that the substring search includes the variable source! So I think you're definitely heading in the right direction here.
2. I would personally have "Group by study" be the default, but I'm very curious what other Dug folks think -- I wonder if we could get you to demo this PR at one of our Tuesday 3pm meetings or at another time? I think we could get a lot of useful feedback on the UI that way.
3. Ideally the counts in the boxes (e.g. "HEAL Studies (4)") should change when you change the histogram -- but maybe that's too complicated? I like how the text above it changes from "Showing X of Y variables" to "Showing X of Y studies" when you turn on "Group by study" mode -- maybe the box could also change from "HEAL Studies (4 studies)" to "HEAL Studies (10 variables)" or something?
4. These results will be a lot more readable if we were to include the form/module name in these search results or if we had a study page to link to, but we don't have either yet :( But someday!

frostyfan109:
> `RUN npm clean-install `is causing the builds to fail, just fyi, is this a custom command or do you mean `RUN npm ci`

npm ci is just an alias of clean-install as far as I know. I think we just need to rebuild the package lock

frostyfan109:
Regarding fuzziness, we were by default using 1 LD if < 5 chars, otherwise 2 LD. I've fixed it to use 1 edit distance (LD) regardless of term length.

src/components/search/results/variable-view-layout/variable-view-layout.js
```diff
@@ -1,34 +1,33 @@
 import { Fragment } from 'react'
 import { Spin, Typography, Empty } from 'antd'
 import { VariableSearchResults } from './'
+import { VariableSearchResults2 } from './variable-results-new'
```
hina-shah, line 4: Can we replace this with VariableSearchResults? Basically cleanup so that we don't have two files for variable-results.

src/components/search/results/variable-view-layout/variable-view-layout.js
```diff
@@ -1,34 +1,33 @@
...
+                    { error.message ? (
+                        <span style={{ marginTop: -144, padding: "0 6px" }}>{ error.message }</span>
+                    ) : !isLoadingVariableResults && totalVariableResults === 0 ? (
+                        <Empty style={{ marginTop: -24 }} description={
+                            <Text type="secondary">No results were found for &quot;{ query }&quot;</Text>
+                        } />
+                        // <span style={{ marginTop: -144, padding: "0 6px" }}>No results found</span>
+                    ) : null }
+                    { variableResults.length > 0 && <ResultsHeader variables={ true } />}
+                    <VariableSearchResults2 />
```
hina-shah, line 29: Replace with VariableSearchResults

src/components/search/results/variable-view-layout/variable-results-new.tsx
```diff
@@ -0,0 +1,647 @@
...
+                        } />
+                    ) }
+                </Space>
+            </InfiniteScroll>
+            <BackTop />
+        </div>
+    )
+}
+
+export const VariableSearchResults2 = () => {
```
hina-shah, line 636: Replace this with VariableSearchResults

src/components/search/results/variable-view-layout/variable-results-new.tsx
hina-shah, line 1: This file is too big for my liking. Is it possible to split into a few files? There seem to be study related components, and variable related components. Maybe split them up? and export the VariableSearchResults with relevant things.?

hina-shah (changes requested):

hina-shah:
Thanks for all the work @frostyfan109 ! A few cleanup things, but other than that I think we should be okay.

hina-shah:
> 
> (3) That's an oversight on my part, I'll change it so the counts reflect the amounts actually being displayed.
Was this implemented?

src/components/search/results/variable-view-layout/variable-results-new.tsx
frostyfan109: will do, I didn't really clean up this branch at all so thanks for catching

frostyfan109:
> > (3) That's an oversight on my part, I'll change it so the counts reflect the amounts actually being displayed.
> > Was this implemented?

I'm not sure, but I don't see it mentioned in newer commits so probably not. I'll check

frostyfan109:
@hina-shah can you approve?

hina-shah (approved):

PR #316, opened 2024-10-11
Adding Develop changes to main

hina-shah:
This PR adds the following features:

- Change explanation tab to have better wording
- Add a user guide to the search UI
- Variable view was re-written to be able to have consistent color over searches.
- Normalizing scores in search results to make them more legible
- In search UI add studies associated with the CDEs in CDE tab of concept card
- UI bug fixes


frostyfan109 (approved):

gaurav:
Some comments:
- [ ] This may not be feasible before this release, but for the next release could we please increase the delay before the tooltips appear? They are useful but they appear too quickly IMO. Just a 1-2 second delay would make this better IMO.
- [ ] We should get someone from the HEAL team to review the tooltip text -- in particular, I think we could shorten a couple of them, which would be nice. I wonder if it would be helpful to display the tooltip text at the top of the box instead of as a pop-up.

.eslintrc.json
```diff
@@ -18,14 +18,18 @@
         "react/prop-types": "off",
         "react/react-in-jsx-scope": "off",
         "react/display-name": "off",
+        "prefer-const": "warn",
```
gaurav, line 21: I think we should open a ticket to turn these rules back on when we have the time -- they seem useful to me, and in the long term I think should be kept on unless we can make a strong argument that they are not necessary.

gaurav:
- [ ] Note that the tour doesn't work locally: clicking "Take a tour" changes us to the help page, so on clicking the first "Next" button the application crashes (presumably because it can't find the search text field). But it works fine on https://heal-dev.apps.renci.org/, so this is either a configuration issue or some bug that only appears when run using `npm run start`.
- [ ] Also, if you don't set brand to `heal`, the "Take a tour" button still appears, but doesn't work. Ideally, it should not appear.

src/components/search/concept-modal/concept-modal.js
```diff
@@ -98,12 +99,78 @@ export const ConceptModalBody = ({ result }) => {
...
-    'kgs':          { title: 'Knowledge Graphs', icon: <KnowledgeGraphsIcon />,  content: <KnowledgeGraphsTab graphs={ graphs } />, },
-    'explanation':  { title: 'Explanation',      icon: <ExplanationIcon />,      content: <ExplanationTab result={ result } /> },
-    'tranql':       { title: 'TranQL',           icon: <TranQLIcon />,           content: <TranQLTab result={ result } graphs = { graphs } /> }
+    'overview': {
+      title: 'Overview',
+      icon: <OverviewIcon />,
+      content: <OverviewTab result={ result } />,
+      tooltip: <div>
+        The Overview tab provides a brief description of the concept.
+        Hovering over the <InfoCircleOutlined style={{ color: "rgba(0, 0, 0, 0.45)", margin: "0 2px" }} />
```
gaurav, line 108: This is a little unclear, because the InfoCircleOutlined isn't actually in the Overview tab but _above_ the Overview tab (as part of the concept model's title).

src/components/search/concept-modal/concept-modal.js
```diff
@@ -98,12 +99,78 @@ export const ConceptModalBody = ({ result }) => {
...
+        &nbsp;besides the concept name displays the corresponding ontological
+        identifier, and where applicable, a link to the entry on the ontology&apos;s website.
+      </div>
+    },
+    'studies': {
+      title: studyTitle,
+      icon: <StudiesIcon />,
+      content: <StudiesTab studies={ studies } />,
+      tooltip: <div>
+        The Studies tab displays studies referencing or researching the concept.
```
gaurav, line 118:
```suggestion
        The Studies tab displays studies whose data dictionaries reference this concept.
```

src/components/search/concept-modal/concept-modal.js
```diff
@@ -98,12 +99,78 @@ export const ConceptModalBody = ({ result }) => {
...
+        Studies are grouped into { studySources.length } categories:&nbsp;
+        { studySources.map((source, i) => (
```
gaurav, line 120: Ooo, I really like this!

src/components/search/concept-modal/concept-modal.js
```diff
@@ -98,12 +99,78 @@ export const ConceptModalBody = ({ result }) => {
...
+      title: cdeTitle,
+      icon: <CdesIcon />,
+      content: <CdesTab cdes={ cdes } cdeRelatedConcepts={ cdeRelatedConcepts } cdeRelatedStudies={cdeRelatedStudies} loading={ cdesLoading } />,
+      tooltip: <div>
+        The CDEs tab displays{ context.brand === "heal" ? " HEAL-approved" : "" } common data elements (CDEs)
+        associated with the concept. A CDE is a standardized question used across studies and clinical
+        trials with a precisely defined set of permissible responses to ensure consistent data collection.
+        {context.brand === "heal" ? <span>
+          &nbsp;The <a href="https://heal.nih.gov/data/common-data-elements" target="_blank" rel="noopener noreferrer">
+            HEAL CDE program
```
gaurav, line 142:
```suggestion
            HEAL CDE program webpage
```

gaurav:
- [ ] Rather than having the "Take a tour" bar visible on the right at all times (where it can obscure content) I would have it either appear as an additional tab in the top right corner (i.e. "Search", "Take a tour", "Get Help"), or possibly only have it appear on the main search page beside the search bar (i.e. "Search", "Take a tour", "Concept | Variable").
- [ ] I thought we decided to remove the "ROBOKOP" tab on the concept modal?

src/components/search/concept-modal/concept-modal.js
```diff
@@ -98,12 +99,78 @@ export const ConceptModalBody = ({ result }) => {
...
+          </a> provides further information.
+        </span> : null}
+      </div>
+    },
+    'kgs': {
+      title: 'Knowledge Graphs',
+      icon: <KnowledgeGraphsIcon />, 
+      content: <KnowledgeGraphsTab graphs={ graphs } />,
+      tooltip: <div>
+        The Knowledge Graphs tab displays the <a href="https://www.oxfordsemantic.tech/faqs/what-is-a-knowledge-graph" target="_blank" rel="noopener noreferrer">
```
gaurav, line 152: I like the idea of linking to a webpage about knowledge graphs, but I would either link to the Wikipedia page (https://en.wikipedia.org/wiki/Knowledge_graph) or the Neo4J page (https://neo4j.com/blog/what-is-knowledge-graph/), which I think are better. Alternatively, you could change this line to:

```suggestion
        The Knowledge Graphs tab displays relevant edges from the <a href="https://robokop.renci.org/" target="_blank" rel="noopener noreferrer">ROBOKOP Knowledge Graph</a>
```

src/components/search/concept-modal/concept-modal.js
```diff
@@ -98,12 +99,78 @@ export const ConceptModalBody = ({ result }) => {
...
+          knowledge graph
+        </a> portion connected to the concept and containing the search terms. This section highlights
```
gaurav, line 154:
```suggestion
        </a> portion connected to the concept and containing your search terms. This section highlights
```

src/components/search/concept-modal/concept-modal.js
```diff
@@ -98,12 +99,78 @@ export const ConceptModalBody = ({ result }) => {
...
+        potential interesting knowledge graph relationships and shows terms (e.g., synonyms) that
```
gaurav, line 155: Do we display synonyms on the knowledge graph tab?

gaurav: Oh, I think I see what you mean: we don't display synonyms, but we do display related concepts (with information on what the relation is, e.g. "[Hepatic fibrosis](http://purl.obolibrary.org/obo/HP_0001395) ---[subclass of]--> [liver disorder](http://purl.obolibrary.org/obo/MONDO_0005154)". So I would simplify this to:

```suggestion
        potential interesting knowledge graph relationships and shows related terms.
```

src/components/search/concept-modal/concept-modal.js
```diff
@@ -221,10 +322,21 @@ export const ConceptModalBody = ({ result }) => {
...
             }
           }))
+          await Promise.all(cdeIds.map(async (cdeId, i) => {
+            try {
+              relatedStudies[cdeId] = await loadRelatedStudies(cdeId)
+            } catch (e) {
+              // Here, we explicitly want to halt execution and forward this error to the outer handler
+              // if a related concept request was aborted, because we now have stale data and don't want to
+              // update state with it.
+              if (e.name === "CanceledError" || e.name === "AbortError") throw e
```
gaurav, line 332: Are errors other than these two silently ignored?

frostyfan109:
> Some comments:
> 
> * [ ]  This may not be feasible before this release, but for the next release could we please increase the delay before the tooltips appear? They are useful but they appear too quickly IMO. Just a 1-2 second delay would make this better IMO. ([This might be how](https://github.com/floating-ui/floating-ui/discussions/1442#discussioncomment-2800095)).
> * [ ]  We should get someone from the HEAL team to review the tooltip text -- in particular, I think we could shorten a couple of them, which would be nice. I wonder if it would be helpful to display the tooltip text at the top of the box instead of as a pop-up.

(1) yes this can be done, I agree this is a good suggestion. FWIW we do not use floating-ui except as a dependency of shepherd itself, the tooltips are implemented using antd/rc-tooltip.
(2) Agree, I've specifically suggested providing a definition/explanation of a "CDE" in the CDE tab, and there's no reason something similar could not be done for others. I personally don't think the tooltips are a great solution, so if we could get the text short enough for all of them, I would prefer embedding it at the top of the tabs instead.

src/components/search/concept-modal/tabs/explanation.js
```diff
@@ -172,10 +182,10 @@ export const ExplanationTab = ({ result }) => {
                         fontSize: 15,
                         fontWeight: 500
                     }}>
-                        Score breakdown
+                        Explanation for this concept&apos;s relation to search term
```
gaurav, line 185:
```suggestion
                        Explanation for this concept&apos;s relation to your search
```

src/contexts/tour-context/context.tsx
```diff
@@ -0,0 +1,876 @@
...
+                attachTo: {
+                    // We want to mount to a sizeless element so that nothing is "highlighted",
+                    // i.e., everything is backdrop except the floating tour popup.
+                    element: "head"
+                },
+                title: "Welcome to HEAL Semantic Search (HSS)",
+                text: renderToStaticMarkup(
+                    <div>
+                        HSS is an open-access search engine for exploring the HEAL research landscape.<br /><br />
+                        HSS is not the search you use every day - it uses linked knowledge to return concepts related
```
gaurav, line 204: The "HSS is not the search you use every day" line annoys me, because Google Search (and, I'm guessing, Bing and DuckDuckGo and all the others) have a lot of semantic search features these days! But I guess this is the approved text from the HEAL Stewards.

src/contexts/tour-context/context.tsx
```diff
@@ -0,0 +1,876 @@
...
+                    middleware: [
+                        tooltipOffset({ mainAxis: -100 })
+                    ]
+                },
+                scrollTo: false,
+                title: "Exploring concept information",
+                text: renderToStaticMarkup(
+                    <div>
+                        You can click on each of the tabs in the side bar for more detailed information, such as 
+                        associated studies, CDEs (common data elements), knowledge graphs, and an explanation
```
gaurav, line 495:
```suggestion
                        associated studies, common data elements (CDEs), knowledge graphs, and an explanation
```

src/components/search/concept-modal/concept-modal.js
frostyfan109: They aren't "ignored", as in, the line immediately following this (333) handles the error. A null value indicates to components downstream that an error occurred. As the comment explains, a request cancellation is explicitly bubbled up  to avoid overwriting state with stale data.

src/contexts/tour-context/context.tsx
```diff
@@ -0,0 +1,876 @@
...
+                id: "main.search.variable-view.study-list",
+                path: searchPathPattern,
+                attachTo: {
+                    element: studyListItemDomMask.selector,
+                    on: "bottom"
+                },
+                scrollToHandler: () => scrollIntoViewIfNeeded(studyListItemDomMask.originalSelector),
+                text: renderToStaticMarkup(
+                    <div>
+                        You can also click to expand each study to view the study variables associated
```
gaurav, line 729: Do we want to highlight that you can now filter variable search results by data source (e.g. "HEAL Studies (284)", "HEAL Research Programs (74)", "Non-HEAL Studies (5441)") in the tour? Maybe in a future release?

gaurav:
- [ ] Click on "Take a tour" from the variable search page gets as far as the "Searching" page, and then stops working.

src/components/search/concept-modal/concept-modal.js
frostyfan109: I'm not sure, perhaps occasionally but in the vast majority of cases, no. The text here I've been told was reviewed and approved by multiple people, so I would assume it's correct but maybe not.

src/contexts/tour-context/context.tsx
```diff
@@ -0,0 +1,876 @@
...
+                path: searchPathPattern,
+                attachTo: {
+                    element: supportNavDomMask.selector,
+                    on: "bottom"
+                },
+                scrollToHandler: () => scrollIntoViewIfNeeded(supportNavDomMask.originalSelector),
+                title: "Conclusion",
+                text: renderToStaticMarkup(
+                    <div>
+                        This concludes the introductory tour of HSS. For more detailed information,
```
gaurav, line 763:
```suggestion
                        This concludes the introductory tour of HSS. For more detailed information, click on the "Get Help" button to
```

gaurav (approved):
All these changes look great -- especially the variable search fixups! I've left a lot of comments, but I suspect most of them will either turn out to be unnecessary or will need to be turned into issues so we can fix them in a future version. But they are all very minor and can be ignored: great work everybody!

.eslintrc.json
frostyfan109: A review of them would not be terrible. But they haven't been disabled carelessly or without thought.

As far as I know, I'm the only developer on this project that uses typescript. React/prop-types without typescript is an archaic rule; you'll never really see explicit proptypes definitions beyond like React 15. Without typescript, this rule requires explicit proptypes.

React-in-jsx-scope is only relevant to projects running React <17.

Prefer-const is generally a good rule. I think I had to set it to warn since a lot of existing code violates it. It could maybe be turned back on.

React/display-name is only for debugging.

src/components/search/concept-modal/concept-modal.js
frostyfan109: true, not sure this belongs here. I think it's more so filler because there's not much to say about an overview?

src/components/search/concept-modal/concept-modal.js
frostyfan109: is it important/relevant to use such a technical term here in a tooltip? I think the purpose is more to offer users a brief explanation of a tab's purpose, which this could be counterproductive to and just confuse them more.

src/components/search/concept-modal/concept-modal.js
frostyfan109: maybe overly verbose. it is a link (so implicitly a webpage) to a page titled "HEAL CDE Program"

src/contexts/tour-context/context.tsx
frostyfan109: I don't really think it makes sense to tell them to "click on the get help button" when the "User Guide" text itself is a link and immediately after, it explains they can navigate to the support page.

frostyfan109: I think the wording should be changed from support page to get help now that this has been renamed, though.

frostyfan109:
> * [ ]  Click on "Take a tour" from the variable search page gets as far as the "Searching" page, and then stops working.

if you can help me reproduce over slack it'd be helpful, though i suspect this may just be a local thing for you. did you reinstall your node_modules when running?

src/components/search/concept-modal/concept-modal.js
hina-shah: I agree with Griffin. This has to be in as laymen terms as possible. People don't know what a data dictionary is as it is!

src/components/search/concept-modal/concept-modal.js
hina-shah: Agreed

hina-shah:
> Some comments:
> 
> * [ ]  This may not be feasible before this release, but for the next release could we please increase the delay before the tooltips appear? They are useful but they appear too quickly IMO. Just a 1-2 second delay would make this better IMO. ([This might be how](https://github.com/floating-ui/floating-ui/discussions/1442#discussioncomment-2800095)).
> * [ ]  We should get someone from the HEAL team to review the tooltip text -- in particular, I think we could shorten a couple of them, which would be nice. I wonder if it would be helpful to display the tooltip text at the top of the box instead of as a pop-up.

I think the tool tip text was reviewed before it was added to the user guide.

src/components/search/concept-modal/concept-modal.js
hina-shah: Added fix.

src/contexts/tour-context/context.tsx
hina-shah: Yep.

hina-shah:
> * [ ]  Click on "Take a tour" from the variable search page gets as far as the "Searching" page, and then stops working.

Is this a bug @frostyfan109 ?

PR #323, opened 2024-10-25
Adding config options for resources on support page, and enable guided tour

hina-shah:
Configuration options to add video URLs under a Resources section on support page.
Also adding a "flag" to enable guided tour. This should be accompanied with the corresponding ui-chart PR:https://github.com/helxplatform/ui-chart/pull/40


YaphetKG (commented):
LGTM! but trivvy scans would be awesome if they passed.

YaphetKG (approved):

PR #332, opened 2025-11-12
CILogon  and Dex implementation

PhillipsOwen:
Associated to MISM-23 ticket.

Adding CILogon and Dex authentication sources.


frostyfan109 (approved):
Looks good. Note that UI signup page functionality will not work with these new redirect-based SSO providers.

PR #335, opened 2026-04-01
Add custom branding for AI sandbox

frostyfan109:
- Navbar theming
- Hero banner image
- Recolored UNC SSO button


src/views/workspaces/login/sso/unc-sso.js
```diff
@@ -2,28 +2,40 @@ import { useMemo } from 'react'
...
 
+    const theme = useMemo(() => (
+        context.brand === 'ai_sandbox' ? {
+            foreground: '#fff',
+            background: '#06667d'
+        } : {
+            foreground: '#fff',
+            background: '#57a0d3'
+        }
+    ), [context])
```
Hoid, line 20: Does `context` change basically every time here? If so, `useMemo` is probably not helping you, and may be even hurting you since the amount of work you're doing inside it is so cheap. Not a huge deal but could be made better by just changing this to `[context.brand]` I think right?

frostyfan109: No, context typically only is set once for a session (page load), yeah we could put context.brand but it wouldn't make any difference.

Hoid: Ah okay. Makes sense. I'll approve then!

Hoid (approved):

PR #337, opened 2026-06-01
update old whitelisting detection code to use new redirect param

frostyfan109:
Code was still checking for redirect to /login_whitelist, but appstore now returns a redirect with query param `?whitelist_required` instead after cilogon changes.


PhillipsOwen (commented):
LGTM

PhillipsOwen (approved):

